PR #42, opened 2020-09-18
[#17] Added api-spec to the project and the maven plugin to generate server boilerplate

### LINKED ISSUE #17: Create Swagger API specification
Create a (first draft) Swagger api specification file for the following api endpoints
- CRUD endpoint for sender
- CRUD endpoint for receiver
- Endpoint for getting a list of all registered senders
- Endpoint for getting a list of all registered receivers

Swagger file should exist somewhere in the project (version control)

Information we are likely to store for senders and/or receivers:

- IP address
- Port
- Nickname/Display name: A human-readable identifier that can be customized by the user
- Device name/serial number: A human-readable identifier chosen by the device that cannot be customized, such as the serial number
- Some sort UUID (to used when getting/updating/deleting)
- Owner (to be implemented later when we add user management)
- Will ask the PM if there's anything else that comes to mind

FGRCL:


service/switchboardapp/pom.xml
```diff
@@ -55,6 +58,19 @@
...
+			<artifactId>openapi-generator</artifactId>
+			<version>${openapi-generator-version}</version>
+		</dependency>
+
+		<!--TODO remove this dependency when org.openapitools.openapi-generator releases version 5.0.0-->
+		<dependency>
+			<groupId>org.openapitools</groupId>
+			<artifactId>jackson-databind-nullable</artifactId>
+			<version>${openapi-jackson-nullable-version}</version>
+		</dependency>
```
FGRCL, line 73: Before anyone asks about this, version 4.3.1 of the open api generator adds an usused import for this library in the models. It is a known bug which is tracked here https://github.com/OpenAPITools/openapi-generator/issues/5690. I personally tested with version 5.0.0-beta2 and the bug is fixed in that version, but I'd rather not depend on a beta version of a library. Another option would be to downgrade to an earlier version where the bug does not exist, but then we risk missing more recent features/bug fixes.

service/switchboardapp/contracts/api-spec.yaml
```diff
@@ -0,0 +1,234 @@
...
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateEncoderRequest'
+      responses:
+        '200':
+          description: Successful creation the encoder
+  /Encoders/{uuid}:
```
sebhani, line 39: Have we decided on using universally unique identifier (uuid)? Why not just an id?

FGRCL: That's pretty much just what I've been doing from experience. Another idea was to use the database IDs, but I figured that would be error prone. Do you have any other ideas?

felixlapierre (approved):
Rather than predict the API, we can come back and change this spec as we go - since nothing else depends on it yet. LGTM for now.

mohhef (approved):
Overall the spec conforms to our initial decisions, will be modified thought the project.

PR #43, opened 2020-09-18
[#7] Set up React base app in Frontend folder

### LINKED ISSUE #7: Create basic frontend project structure
In the repository, create a folder named `frontend`. Inside that folder, create a basic frontend that will become our admin panel.

Start with React, if we have trouble then we`ll switch to plain html/css/js

jun-loke:
## PR

**Issue**: https://github.com/felixlapierre/switchboard/issues/7

**Task description**: 
 
Created base React application in **frontend** folder

## Testing

**Test coverage:** 
N/A

## Additional notes

**Note to reviewers**

To run this application, go into the `frontend`  folder and run `npm start` for a development build (not optimized) 
or `npm run build` for a production build


FGRCL:
I'm going to suggest we send the "note to reviewers" in the general slack channel once this PR is merged

frontend/.gitignore
```diff
@@ -0,0 +1,23 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
FGRCL, line 1: Should we add gitignore lines for IDE files and directories? I'm thinking VS code, any suggestions?

frontend/README.md
```diff
@@ -0,0 +1,68 @@
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
```
FGRCL, line 1: Is this a default readme from create-react-app?

frontend/public/index.html
```diff
@@ -0,0 +1,43 @@
...
+    <!--
+      Notice the use of %PUBLIC_URL% in the tags above.
+      It will be replaced with the URL of the `public` folder during the build.
+      Only files inside the `public` folder can be referenced from the HTML.
+
+      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
+      work correctly both with client-side routing and a non-root public URL.
+      Learn how to configure a non-root public URL by running `npm run build`.
+    -->
+    <title>React App</title>
```
FGRCL: Do we need to change this title to switchboard?

frontend/public/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
+  "short_name": "React App",
```
FGRCL: We might want to change these two names?

FGRCL:
Did we choose a react component library yet?

frontend/README.md
jun-loke: yes, it's the default

frontend/public/manifest.json
jun-loke: Sure, did you have any ideas in mind?

frontend/.gitignore
jun-loke: That would be good. But probably more important on a project wide basis rather than just within the frontend folder

jun-loke:
> Did we choose a react component library yet? 
@FGRCL 

Not yet. We were waiting for Sharon's mockups which were received yesterday.
Looks like Bootstrap or Material UI kind of esthetic. I'm doing some research into which one would be better but I'm leaning towards Material.

frontend/public/index.html
jun-loke: Changed in 751de33

frontend/public/manifest.json
FGRCL: What do you think about `Switchboard` and `Switchboard Application`?

frontend/.gitignore
FGRCL: I was expecting that we'd open the project in the root of the frontend folder instead of the root of the repo, in which case it'd be safe to add it to this .gitignore.

mohhef:
I think its better to use Yarn as a package manager for react. If we choose either yarn or npm we need to make sure all dependencies are installed only using one or the other.

FGRCL:
@mohhef 
> I think its better to use Yarn as a package manager for react. If we choose either yarn or npm we need to make sure all dependencies are installed only using one or the other.

I've never used Yarn before. I'm curious, why do you like yarn more than npm?

mohhef:
@FGRCL 
> @mohhef
> 
> > I think its better to use Yarn as a package manager for react. If we choose either yarn or npm we need to make sure all dependencies are installed only using one or the other.
> 
> I never used Yarn before. I'm curious, why do you like yarn more than npm?

Mainly because we had a lot of package management issues using npm in 390(react native project) so we switched to yarn. It's as simple as using npm. Both react and yarn are developed by FB which makes it more optimized.

felixlapierre:
> 
>@mohhef
> Mainly because we had a lot of package management issues using npm in 390(react native project) so we switched to yarn. It's as simple as using npm. Both react and yarn are developed by FB which makes it more optimized.

Can you explain the issue you had or link the PR for me? I've never had any issues like that with npm before.

mohhef:
> > @mohhef
> > Mainly because we had a lot of package management issues using npm in 390(react native project) so we switched to yarn. It's as simple as using npm. Both react and yarn are developed by FB which makes it more optimized.
> 
> Can you explain the issue you had or link the PR for me? I've never had any issues like that with npm before.

If I remember correctly npm would cause some performance issues or something with versioning for people who were initially setting up the base of the project. Can't find a pr for it. Since people are mostly familiar with npm, won't make much of a difference anyway, especially since I just read newer versions of npm managed the gap of the performance differences.

package-lock.json
```diff
@@ -0,0 +1,3 @@
+{
+  "lockfileVersion": 1
+}
```
felixlapierre: Why is there a package-lock file in the root? Can we get rid of it?

frontend/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "frontend",
```
felixlapierre: Change the name to something like "Switchboard" or "Switchboard App" or "Switchboard Frontend"

frontend/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "@testing-library/user-event": "^7.2.1",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1",
+    "react-scripts": "3.4.3"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject"
```
felixlapierre: React does this really annoying thing where `react-scripts test` only runs tests on files that have changes since the last commit.
Can you add a script that runs all the tests regardless? Just add the following line
```
    "test-all": "react-scripts test --watchAll=false"
```

felixlapierre (changes requested):
Good job y'all! I was able to build and run locally and everything worked fine, tests all pass.

I just have a couple of minor changes i'd like you to make before we merge, thx 😄

package-lock.json
gdzooks: hmmm... considering there's a package-lock with actual contents in the frontend folder, i think it's safe to remove this

frontend/package.json
jun-loke: Fixed in abe57fb

frontend/package.json
jun-loke: added in 01db205
Typo fixed in b937456

package-lock.json
jun-loke: Removed in 0088853

FGRCL (dismissed):

frontend/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "switchboard": "switchboard frontend",
```
felixlapierre: Little typo
```suggestion
  "name": "switchboard frontend",
```

jun-loke: There's actually a bit more than that. the name needs to be a single word, no spaces.
rules about it are defined here: https://docs.npmjs.com/creating-a-package-json-file
I'm gonna elect for snake case, unless there's a standard for it

jun-loke: Amended in 76b3bd2

felixlapierre (approved):

FGRCL (approved):

PR #62, opened 2020-09-30
[#6/14/15/21] Endpoints for receiver and sender

### LINKED ISSUE #6: Service: Endpoint for getting a list of senders

### LINKED ISSUE #14: Service: Endpoint for registering a sender
Also includes setting up the database (mysql, jdbc or hibernate, etc)

### LINKED ISSUE #15: Service: Endpoint for registering a receiver

### LINKED ISSUE #21: Service: Endpoint for getting a list of receivers

### LINKED ISSUE #22: Service: Endpoints for getting one sender/receiver

### LINKED ISSUE #61: Post request to/encoder results in an internal server error
**OS:** Windows 10

**Steps to reproduce:** 

Make a post request to /encoder using the following body:
{
    "serialNumber": 2
}

**Expected result:**

The body should be populated in the database and the server should return a successful response.

**Actual result:**

Return of an internal server error.

**Visual proof:**

mohhef:
# General info
Connection to MySQL: URL/password/username are included in system.properties
Use postman to test post and get requests

**Issue number**:  6/14/15/21

**Task description**: 

- Establish database connection to MySQL
- Map entities to the database
- Create post and get methods for device/decoder/encoder(retrieveAllX, retrieveX, createX)

**Note to reviewers**:
This pull request does not directly implement the API spec because of structural changes, API spec to be implemented
Component tests to be added to the HTTP requests

Reasons for having endpoints for Device:
- To create a device without getting assigned to either an encoder or a decoder
- To deassign a device from an encoder then reassign it to a decoder without deleting all its attributes and visa versa
- To query getting all the devices without queries 2 individual tables

Access api used by the appliction(not yaml): http://localhost:8080/swagger-ui/index.html?configUrl=/v3/api-docs/swagger-config
The swagger-ui should match the yaml when openApi is implemented properly, however you can still use swagger-ui endpoint for testing purposes


service/switchboardapp/src/main/java/com/switchboard/app/controller/DeviceController.java
```diff
@@ -0,0 +1,62 @@
...
+    @Autowired
+    DeviceDaoImpl service;
+
+    @GetMapping("/device")
+    public List<Device> retrieveAllDevices(){
+        return service.getDevices();
+    }
+
+    @GetMapping("/device/{serialNumber}")
+    public EntityModel<Device> retrieveDevice(@PathVariable long serialNumber){
```
felixlapierre: As misleading as it is, serial numbers can contain letters, so we have to store them as strings :/

service/switchboardapp/src/main/java/com/switchboard/app/controller/EncoderController.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Optional;
+
+@RestController
+public class EncoderController {
+
+    private Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    @Autowired
+    EncoderDaoImpl encoderService;
+//
```
felixlapierre: Remove comment

service/switchboardapp/src/main/java/com/switchboard/app/controller/EncoderController.java
```diff
@@ -0,0 +1,51 @@
...
+    public List<Encoder> retrieveAllEncoders(){
+        return encoderService.getEncoders();
+    }
+
+    @PostMapping("/encoder")
+    public ResponseEntity createEncoder(@RequestBody Encoder encoder){
+        Optional<Device> deviceOptional = deviceService.findDevice(encoder.getSerialNumber());
+        encoder.setDevice(deviceOptional.get());
+
+        //  encoderRepository.saveAndFlush(encoder);
```
felixlapierre: Remove commented code

service/switchboardapp/src/main/java/com/switchboard/app/dao/DecoderDao.java
```diff
@@ -0,0 +1,17 @@
...
+import com.switchboard.app.domain.Device;
+import com.switchboard.app.domain.Encoder;
+import org.springframework.stereotype.Repository;
+
+import java.util.List;
+import java.util.Optional;
+
+public interface DecoderDao {
+
+//    Decoder addDao(Decoder decoder);
```
felixlapierre: Remove commented code

service/switchboardapp/src/main/resources/application.properties
```diff
@@ -1 +1,11 @@
+spring.h2.console.enabled=true
+spring.jpa.hibernate.ddl-auto=update
 
+#spring.datasource.url=jdbc:h2:mem:testdb
+#spring.jpa.show-sql=true
```
felixlapierre: Commented properties, do we need to keep them for easily enabling stuff in the future?

service/switchboardapp/src/main/resources/data.sql
```diff
@@ -0,0 +1,4 @@
+insert into device(serial_number, display_name, Status) values (1,'Device1','Running');
+insert into device(serial_number, display_name, Status) values (2,'Device2','Running');
+
+insert into encoder(serial_number) values(1);
```
felixlapierre: Is this just default data to insert into the database?

service/switchboardapp/src/main/java/com/switchboard/app/exceptions/CustomRestExceptionHandler.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
+
+@ControllerAdvice
+@RestController
+public class CustomRestExceptionHandler extends ResponseEntityExceptionHandler {
+
+    @ExceptionHandler(Exception.class)
+    public final ResponseEntity<Object> handleAllExceptions(Exception ex, WebRequest request){
+
+       ExceptionResponse exceptionResponse = new ExceptionResponse(HttpStatus.INTERNAL_SERVER_ERROR, ex.getMessage(), "error occured");
```
felixlapierre: Change message to "An unknown error occurred"

service/switchboardapp/src/main/java/com/switchboard/app/domain/Device.java
```diff
@@ -0,0 +1,30 @@
...
+    @Column(name = "display_name")
+    private String DisplayName;
+
+    private String Status;
+
+    @OneToOne(mappedBy = "device")
+    private Encoder encoder;
+
+    @OneToOne(mappedBy = "device", cascade = CascadeType.ALL, fetch = FetchType.LAZY)
+    private Decoder decoder;
```
felixlapierre: Why does Encoder not have CascadeType.ALL and FetchType.LAZY?

service/switchboardapp/src/main/java/com/switchboard/app/domain/Decoder.java
```diff
@@ -0,0 +1,27 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+import javax.persistence.*;
+
+@Entity
+@AllArgsConstructor
+@NoArgsConstructor
+@Setter@Getter
```
felixlapierre: I suggest replacing @Setter and @Getter with @Data, it will include those two annotations plus ToString, HashCode, and RequiredArgsConstructor

service/switchboardapp/src/main/java/com/switchboard/app/controller/DecoderController.java
```diff
@@ -0,0 +1,48 @@
...
+    @Autowired
+    DecoderDaoImpl decoderService;
+
+    @Autowired
+    DeviceDaoImpl deviceService;
+
+    @GetMapping("/decoder")
+    public List<Decoder> retrieveAllDecoders(){
+        return decoderService.getDecoders();
+    }
```
felixlapierre: I see that we support retrieving a specific device (which includes its encoder and decoder info) but should we support retrieving a specific encoder or decoder too?

service/switchboardapp/src/main/java/com/switchboard/app/domain/Decoder.java
```diff
@@ -0,0 +1,27 @@
...
+@JsonIgnoreProperties({"hibernateLazyIntializer","handler","device"})
+
+public class Decoder {
+
+    @Id
+    @Column(name = "serial_number")
+    private Long serialNumber;
```
felixlapierre: Add @NotNull unless that's already validated through the other annotations. If you do, add @Valid to the create new decoder endpoint.

service/switchboardapp/src/main/java/com/switchboard/app/domain/Device.java
```diff
@@ -0,0 +1,30 @@
...
+@ToString
+@JsonIgnoreProperties({"hibernateLazyIntializer","handler","encoder","decoder"})
+public class Device {
+
+    @Id
+    @Column(name = "serial_number")
+    private Long serialNumber;
+
+    @Column(name = "display_name")
+    private String DisplayName;
```
felixlapierre: Add @NotNull to serial number and display name

service/switchboardapp/src/main/java/com/switchboard/app/domain/Encoder.java
```diff
@@ -0,0 +1,24 @@
...
+@AllArgsConstructor
+@NoArgsConstructor
+@Getter@Setter
+@JsonIgnoreProperties({"hibernateLazyIntializer","handler","device"})
+@ToString
+public class Encoder {
+
+    @Id
+    @Column(name = "serial_number")
+    private Long serialNumber;
```
felixlapierre: Add @NotNull unless that's already validated by the other annotations. If you do, add @Valid to the create encoder endpoint Encoder argument.

felixlapierre (changes requested):
Looking good!! Have some minor suggestions + feedback, and one major change: we have to change Serial Number to a string instead of a long since it can contain letters (confusing, i know)

service/switchboardapp/contracts/api-spec.yaml
```diff
@@ -11,7 +11,7 @@ servers:
 paths:
   /Encoders:
     get:
-      operationId: getEncoders
+      operationId: retrieveAllEncoders
```
FGRCL, line 46: Why did we change this name?

service/switchboardapp/pom.xml
```diff
@@ -40,13 +40,26 @@
 			<scope>runtime</scope>
 			<optional>true</optional>
 		</dependency>
-
+		<!-- https://mvnrepository.com/artifact/com.h2database/h2 -->
```
FGRCL: I'm not sure what this comment is trying to specify

service/switchboardapp/pom.xml
```diff
@@ -40,13 +40,26 @@
...
-
+		<dependency>
+			<groupId>mysql</groupId>
+			<artifactId>mysql-connector-java</artifactId>
+			<scope>runtime</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.hateoas</groupId>
+			<artifactId>spring-hateoas</artifactId>
+			<version>1.1.2.RELEASE</version>
```
FGRCL: let's define this versions in the <properties> section.

service/switchboardapp/pom.xml
```diff
@@ -58,7 +71,10 @@
 				</exclusion>
 			</exclusions>
 		</dependency>
-
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-data-jpa</artifactId>
+        </dependency>
```
FGRCL, line 86: This dependency is not on the same indent level as the other dependencies.

service/switchboardapp/src/main/java/com/switchboard/app/SwitchboardApplication.java
```diff
@@ -8,6 +8,7 @@
 
 	public static void main(String[] args) {
 		SpringApplication.run(SwitchboardApplication.class, args);
+
```
FGRCL: I think we can remove this empty line.

service/switchboardapp/src/main/java/com/switchboard/app/controller/DecoderController.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.validation.Valid;
+import java.net.URI;
+import java.util.List;
+import java.util.Optional;
+
+import static org.springframework.hateoas.server.mvc.WebMvcLinkBuilder.linkTo;
+import static org.springframework.hateoas.server.mvc.WebMvcLinkBuilder.methodOn;
+
+@RestController
+public class DecoderController {
```
FGRCL, line 25: This controller does not inherit the DecodersApi interface

service/switchboardapp/src/main/java/com/switchboard/app/controller/DecoderController.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Autowired
+    DecoderDaoImpl decoderService;
+
+    @Autowired
+    DeviceDaoImpl deviceService;
```
FGRCL, line 31: I think we should define these as private final variables and use Lombok to generate the constructors. I think the dependencies will be easier to mock for unit testing that way. Lombok has a little bit of config that needs to be done with maven and Intellij, I've done it on my side so I can show you how to do it.

service/switchboardapp/src/main/java/com/switchboard/app/controller/DecoderController.java
```diff
@@ -0,0 +1,48 @@
...
+    public List<Decoder> retrieveAllDecoders(){
+        return decoderService.getDecoders();
+    }
+
+    @PostMapping("/decoder")
+    public ResponseEntity createDecoder(@RequestBody Decoder decoder){
+        Optional<Device> deviceOptional = deviceService.findDevice(decoder.getSerialNumber());
+        decoder.setDevice(deviceOptional.get());
+
+        Decoder savedDecoder = decoderService.addDecoder(decoder);
```
FGRCL: This logic seems like it belongs more in a repository or DAO

service/switchboardapp/src/main/java/com/switchboard/app/controller/DecoderController.java
```diff
@@ -0,0 +1,48 @@
...
+        URI location = ServletUriComponentsBuilder.fromCurrentRequest().
+                path("/{serialNumber}").buildAndExpand(savedDecoder.getSerialNumber()).toUri();
+
+        return ResponseEntity.created(location).build();
```
FGRCL, line 62: For now, our api-specs only specify that we return a 200 when an object is successfully created.

service/switchboardapp/src/main/java/com/switchboard/app/controller/DeviceController.java
```diff
@@ -0,0 +1,62 @@
...
+import javax.validation.Valid;
+import java.net.URI;
+import java.util.List;
+import java.util.Optional;
+
+import static org.springframework.hateoas.server.mvc.WebMvcLinkBuilder.linkTo;
+import static org.springframework.hateoas.server.mvc.WebMvcLinkBuilder.methodOn;
+
+@RestController
+public class DeviceController implements DecodersApi {
```
FGRCL: Why are we adding this controller? the API never mentions a user facing Device model.

service/switchboardapp/src/main/java/com/switchboard/app/controller/DeviceController.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Autowired
+    DeviceDaoImpl service;
+
+    @GetMapping("/device")
+    public List<Device> retrieveAllDevices(){
+        return service.getDevices();
+    }
+
+    @GetMapping("/device/{serialNumber}")
```
FGRCL: Those mappings aren't part of our API specifications

service/switchboardapp/src/main/java/com/switchboard/app/controller/DeviceController.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Optional;
+
+import static org.springframework.hateoas.server.mvc.WebMvcLinkBuilder.linkTo;
+import static org.springframework.hateoas.server.mvc.WebMvcLinkBuilder.methodOn;
+
+@RestController
+public class DeviceController implements DecodersApi {
+
+    @Autowired
+    DeviceDaoImpl service;
```
FGRCL, line 26: Same comment about Lombok.

service/switchboardapp/src/main/java/com/switchboard/app/controller/EncoderController.java
```diff
@@ -0,0 +1,51 @@
...
+public class EncoderController {
+
+    private Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    @Autowired
+    EncoderDaoImpl encoderService;
+//
+
+    @Autowired
+    DeviceDaoImpl deviceService;
```
FGRCL: Same comment about lombok.

service/switchboardapp/src/main/java/com/switchboard/app/controller/EncoderController.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.web.bind.annotation.*;
+import org.springframework.web.servlet.support.ServletUriComponentsBuilder;
+
+import javax.validation.Valid;
+import java.net.URI;
+import java.util.List;
+import java.util.Optional;
+
+@RestController
+public class EncoderController {
```
FGRCL, line 26: This controller should inherit the EncoderApi interface.

service/switchboardapp/src/main/java/com/switchboard/app/controller/EncoderController.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private Logger logger = LoggerFactory.getLogger(this.getClass());
```
FGRCL: Lombok has a @Slf4j defines this logger for you.

service/switchboardapp/src/main/java/com/switchboard/app/controller/EncoderController.java
FGRCL: same comment about lombok

service/switchboardapp/src/main/java/com/switchboard/app/dao/impl/DecoderDaoImpl.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public Decoder addDecoder(Decoder decoder) {
+        return decoderRepository.save(decoder);
+    }
+
+    @Override
+    public List<Decoder> getDecoders() {
+        return decoderRepository.findAll();
+    }
+}
```
FGRCL: I believe repositories and DAO come from two different design patterns for the persistence layer. I'm going to suggest we move the methods of the DAOs to their corresponding repositories instead and get rid of the DAOs altogether.

service/switchboardapp/src/main/java/com/switchboard/app/domain/Decoder.java
```diff
@@ -0,0 +1,27 @@
...
+
+import javax.persistence.*;
+
+@Entity
+@AllArgsConstructor
+@NoArgsConstructor
+@Setter@Getter
+@JsonIgnoreProperties({"hibernateLazyIntializer","handler","device"})
+
+public class Decoder {
```
FGRCL: I think we should postfix our persistence layer objects with "Entities" to avoid confusion. I also intend on postfixing the API models with "Model".

service/switchboardapp/src/main/java/com/switchboard/app/domain/Decoder.java
FGRCL: Is it safe to use serial numbers as ID? From what I understood we could have multiple clients with the same serial number.

service/switchboardapp/src/main/java/com/switchboard/app/exceptions/DeviceAlreadyExistsException.java
```diff
@@ -0,0 +1,7 @@
+package com.switchboard.app.exceptions;
+
+public class DeviceAlreadyExistsException extends RuntimeException {
+    public DeviceAlreadyExistsException(String s) {
+        super(s);
+    }
```
FGRCL, line 6: With Lombok I think we could get rid of the constructor with @RequiredArgsConstructor

service/switchboardapp/src/main/java/com/switchboard/app/exceptions/DeviceNotFoundException.java
```diff
@@ -0,0 +1,7 @@
+package com.switchboard.app.exceptions;
+
+public class DeviceNotFoundException extends RuntimeException {
+    public DeviceNotFoundException(String s) {
+        super(s);
+    }
```
FGRCL, line 8: Same comment about Lombok and the constructor

service/switchboardapp/src/main/java/com/switchboard/app/exceptions/ExceptionResponse.java
```diff
@@ -0,0 +1,25 @@
...
+@AllArgsConstructor
+@Getter
+public class ExceptionResponse {
+
+    public HttpStatus status;
+    public String message;
+    public List<String> errors;
+
+    public ExceptionResponse(HttpStatus status, String message, String error) {
+        super();
```
FGRCL: This class doesn't inherit anything. I think we can remove this super.

service/switchboardapp/src/main/java/com/switchboard/app/exceptions/ExceptionResponse.java
```diff
@@ -0,0 +1,25 @@
...
+        this.status = status;
+        this.message = message;
+        errors = Arrays.asList(error);
```
FGRCL: Th `Arrays.asList` doesn't seem to be doing much here. When applied on a string, `Arrays.asList` will return a list with a single element which is the string passed. We should change the `error` method parameter to either an array, a String or use varargs.

service/switchboardapp/src/main/resources/data.sql
FGRCL: Is this data added on boot? If that's the case we should make sure it's loaded only when the app is loaded with a test or debug profile.

FGRCL (changes requested):

service/switchboardapp/src/main/java/com/switchboard/app/dao/impl/DecoderDaoImpl.java
FGRCL: I think the DAOs currently do the job of the service layer so we could rename them to that. In any case, the service layer should return their own data objects instead or the persistence layer's data objects.

FGRCL:
Let's add #6  and #21 to this pull request

service/switchboardapp/src/main/java/com/switchboard/app/controller/DeviceController.java
mohhef: fixed

service/switchboardapp/src/main/java/com/switchboard/app/controller/EncoderController.java
mohhef: fixed

service/switchboardapp/src/main/resources/application.properties
mohhef: Ill keep the h2 property for now, in case anybody cant connect to sql they can test the application using h2.

service/switchboardapp/src/main/java/com/switchboard/app/domain/Decoder.java
mohhef: Good  idea

mohhef: fixed

service/switchboardapp/src/main/java/com/switchboard/app/controller/DecoderController.java
mohhef: Yes. we did not do it in this pull request, we'll finish the rest of the concerns and get back to this one.

service/switchboardapp/src/main/resources/data.sql
mohhef: This data is loaded when using h2 in memory database. Ill make that clear!

service/switchboardapp/src/main/java/com/switchboard/app/controller/EncoderController.java
mohhef: fixed

service/switchboardapp/src/main/java/com/switchboard/app/exceptions/CustomRestExceptionHandler.java
mohhef: Fixed

service/switchboardapp/src/main/java/com/switchboard/app/domain/Decoder.java
mohhef: I think @id will fail in the db level if the value is null, adding @NotNull is also a good idea to fail at the method validation level

mohhef: fixed

service/switchboardapp/src/main/java/com/switchboard/app/domain/Device.java
mohhef: fixed

service/switchboardapp/src/main/java/com/switchboard/app/domain/Encoder.java
mohhef: fixed

service/switchboardapp/pom.xml
mohhef: Removed.
Fixed.

service/switchboardapp/pom.xml
mohhef: Fixed

service/switchboardapp/src/main/java/com/switchboard/app/SwitchboardApplication.java
mohhef: Fixed :)

service/switchboardapp/pom.xml
mohhef: On my local its fine but on GitHub you are right, whats causing that, do you have any idea?

service/switchboardapp/src/main/java/com/switchboard/app/exceptions/DeviceAlreadyExistsException.java
mohhef: doesnt seem to work :/

service/switchboardapp/src/main/java/com/switchboard/app/exceptions/DeviceNotFoundException.java
mohhef: doesnt seem to work :/

service/switchboardapp/src/main/java/com/switchboard/app/exceptions/ExceptionResponse.java
mohhef: Changed to string for now

service/switchboardapp/src/main/java/com/switchboard/app/exceptions/ExceptionResponse.java
mohhef: Yes, fixed

service/switchboardapp/src/main/java/com/switchboard/app/controller/EncoderController.java
mohhef: fixed

service/switchboardapp/src/main/java/com/switchboard/app/dao/DecoderDao.java
mohhef: fixed

service/switchboardapp/src/main/java/com/switchboard/app/dao/impl/DecoderDaoImpl.java
mohhef: My intention was that the repository will only contain custom queries that will not be mixed with JPA readily available operations. I however refactored the code, take a look at it and let me know. Now, we will include  the custom queries and the JPA readily available operations in the repository.

service/switchboardapp/src/main/java/com/switchboard/app/controller/DeviceController.java
mohhef: This controller will let the front end fetch all devices wheather they are set to encoder or decoder or not to anything. Assuming that when you first register an encoder or a decoder you have to first register it as a device and then assign it an encoder or a decoder

service/switchboardapp/src/main/java/com/switchboard/app/controller/DecoderController.java
mohhef: why do you think that

service/switchboardapp/src/main/java/com/switchboard/app/domain/Decoder.java
mohhef: fixed

service/switchboardapp/src/main/java/com/switchboard/app/domain/Decoder.java
mohhef: not sure about that, from my understanding serial numbers are unique. We can have multiple ports to a specific serial number

service/switchboardapp/pom.xml
FGRCL: I'm don't have any ideas of the top of my head, but maybe we could avoid issues like this by enforcing a CodeStyle file in IntelliJ for all the team. What do you think about it?

mohhef: Thats a good idea, but i just found out we can install 'save actions' plugin and it will reformat the file, organize imports, etc.. on saving a file(intellij autosaves, might not be ideal to reformat in the middle of coding) or on pressing ctrl+shift+s. Ill try to do that on the pom.xml and see what happens

mohhef: it fixed the issue! codestyle to enforce the styles is still good suggestion to enforce the styles

service/switchboardapp/src/main/java/com/switchboard/app/domain/Device.java
mohhef: Both should not have them, fixed.

service/switchboardapp/contracts/api-spec.yaml
mohhef: i changed the ids  to match the method names we had, i also changed other stuff. I dont know the best conventions for the api-spec document( please advise if theres something wrong), however i made some changes so that it atleast matches our current code. I will implement it this pr if needed

mohhef:
All comments have been addressed except enforcing the yaml script on the implementation, I suggest merging this PR and doing the enforce in the next PR

service/switchboardapp/src/main/java/com/switchboard/app/controller/DecoderController.java
mohhef: fixed

service/switchboardapp/contracts/api-spec.yaml
```diff
@@ -21,7 +53,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/EncoderUuids'
+                $ref: '#/components/schemas/Encoder'
```
FGRCL, line 56: This model is inconsistent with the description of the endpoints. This model contains a single Encoder object while the endpoint should return a list of some kind. The reason I made it return a list of IDs is to reduce the amount of data in case there were a lot of encoders. But let's take care of that in the next PR

felixlapierre (approved):
Looks good to merge for now. We'll have to create tasks to represent the unfinished work of:
- Matching the controllers to the interfaces generated by the api spec
- Write unit and integration tests for this code

FGRCL (approved):

PR #87, opened 2020-10-08
[#84] Add SonarCloud GitHub Action

### LINKED ISSUE #84: Quality Assurance Tools
Run Sonarqube on pull requests using Github Actions

ButlerJAL:
# General info
Add sonarcloud github action

**Issue number**: #84 

**Task description**: 

 - Run Sonarqube on pull requests using Github Actions

# Testing

**Test coverage**: n/a

# Additional notes

**Note to reviewers**:
Must be merged into master for github action to start taking effect.

To view the code quality report you must have an account on SonarCloud. [Sign up for free now](https://sonarcloud.io/sessions/init/github) if it's not already the case!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,36 @@
+name: Build
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Cache SonarCloud packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar
+          restore-keys: ${{ runner.os }}-sonar
+      - name: Cache Maven packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Build and analyze
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        run: mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,5 +33,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-        run: mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
+        run: mvn -B verify -DskipTests org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
         working-directory: ./service/switchboardapp
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,3 +34,4 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
         run: mvn -B verify org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
+        working-directory: ./service/switchboardapp
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,6 @@
+sonar.organization=bean-pod
+sonar.projectKey=bean-pod_switchboard
+
+# relative paths to source directories. More details and properties are described
+# in https://sonarcloud.io/documentation/project-administration/narrowing-the-focus/
+sonar.sources=.
--- a/service/switchboardapp/pom.xml
+++ b/service/switchboardapp/pom.xml
@@ -23,6 +23,9 @@
         <openapi-jackson-nullable-version>0.2.1</openapi-jackson-nullable-version>
         <maven-compiler-plugin-version>3.8.1</maven-compiler-plugin-version>
         <lombok-version>1.18.12</lombok-version>
+        <sonar.projectKey>bean-pod_switchboard</sonar.projectKey>
+        <sonar.organization>bean-pod</sonar.organization>
+        <sonar.host.url>https://sonarcloud.io</sonar.host.url>
     </properties>
 
     <dependencies>


ButlerJAL:
Is there a particular reason that we are not following [the standard directory layout](http://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) for maven projects? I needed to specify that switchboardapp is the working directory. I recommend that we standardize the directory layout to have our pom.xml file at the top level.

Also, I'm not sure why my additions would cause a test failure, but I had to include -DskipTests argument in the maven run command of the build.yml file for the sonar github action to build and analyze successfully. Possibly the contributor who added the spring boot test could determine the failure point?

FGRCL:
> Is there a particular reason that we are not following [the standard directory layout](http://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) for maven projects? I needed to specify that switchboardapp is the working directory. I recommend that we standardize the directory layout to have our pom.xml file at the top level.
> 
> Also, I'm not sure why my additions would cause a test failure, but I had to include -DskipTests argument in the maven run command of the build.yml file for the sonar github action to build and analyze successfully. Possibly the contributor who added the spring boot test could determine the failure point?

In the current state the tests fail if you don't have database running for the back-end to connect to. The app won't load successfully if it is unable to connect to the database. We need to figure out how to have database running in github actions or maybe mock the database during the pipeline. Regardless, I believe @sebhani  is currently investigating integration tests for the back-end, he might have some ideas on how to approach this. My initial thought would be to try and use docker to run a database

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Build and analyze
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        run: mvn -B verify -DskipTests org.sonarsource.scanner.maven:sonar-maven-plugin:sonar
```
FGRCL, line 36: In the github actions documentation, the example workflow file has a `-` before run, it might be required for multiple `run` steps. [the example file](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/introduction-to-github-actions#create-an-example-workflow)

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar
+          restore-keys: ${{ runner.os }}-sonar
+      - name: Cache Maven packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
```
FGRCL, line 31: What do these two steps do? Also I think they might have some warnings when they run looking at [the check log](https://github.com/bean-pod/switchboard/pull/87/checks?check_run_id=1229408306).

.github/workflows/build.yml
FGRCL, line 36: Should we use the maven wrapper in `service/switchboardapp/mvnw` to avoid weird version issues? it might also avoid you the trouble of setting `working-directory` but I'm not sure.

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
+name: Build
```
FGRCL, line 1: Do you have an idea on how we could setup the different stages of our pipeline? Looking at the documentation it seems like we could do either multiple workflows or multiple steps. Also jobs run in parallel, which could be interesting when running unit tests and integration tests for example.

sonar-project.properties
```diff
@@ -0,0 +1,6 @@
+sonar.organization=bean-pod
+sonar.projectKey=bean-pod_switchboard
+
+# relative paths to source directories. More details and properties are described
+# in https://sonarcloud.io/documentation/project-administration/narrowing-the-focus/
+sonar.sources=.
```
FGRCL, line 6: Could we explicitly specify the source code folders here, for the back-end and the front-end? Which would be
`/service/switchboardapp/src` and `frontend/src`. I want to avoid getting noise in the reports if sonarqube starts anaylzing random files like the pom.xml, package-lock.json, and others.

FGRCL (changes requested):

.github/workflows/build.yml
ButlerJAL: The `-` has nothing to do with multiple run steps. Here is the [official YAML Spec](https://yaml.org/spec/1.2/spec.html) if you wish to get in the nitty gritty of things. [The example file](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/introduction-to-github-actions#create-an-example-workflow) is a simple example with one group of actions to check the bats version, so, it makes sense to just have the actions listed in that way. In our case the build job is a little more complex with many groups of actions, so, the compact nested mapping makes sense here.

.github/workflows/build.yml
ButlerJAL: These two steps allow caching dependencies and build outputs to improve workflow execution time. I did not see any warnings related to these two steps in [the check log](https://github.com/bean-pod/switchboard/pull/87/checks?check_run_id=1229408306). You can read on actions/cache [here](https://github.com/actions/cache) to get all the details. The warning I believe you are referring to is related to spring boot.

.github/workflows/build.yml
ButlerJAL: We should use [Maven's standard directory layout](http://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html).

sonar-project.properties
ButlerJAL: Yes we can. As already stated, our repository should follow [Maven's standard directory layout](http://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) rather than specifying folders.

.github/workflows/build.yml
ButlerJAL: It's possible to have multiple .yml files in `.github/workflows` for multiple workflows.

ButlerJAL: Here is a good reference to understand the [workflow syntax for github actions](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions).

ButlerJAL:
> > Is there a particular reason that we are not following [the standard directory layout](http://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) for maven projects? I needed to specify that switchboardapp is the working directory. I recommend that we standardize the directory layout to have our pom.xml file at the top level.
> > Also, I'm not sure why my additions would cause a test failure, but I had to include -DskipTests argument in the maven run command of the build.yml file for the sonar github action to build and analyze successfully. Possibly the contributor who added the spring boot test could determine the failure point?
> 
> In the current state the tests fail if you don't have database running for the back-end to connect to. The app won't load successfully if it is unable to connect to the database. We need to figure out how to have database running in github actions or maybe mock the database during the pipeline. Regardless, I believe @sebhani is currently investigating integration tests for the back-end, he might have some ideas on how to approach this. My initial thought would be to try and use docker to run a database

Good to know

felixlapierre:
> In the current state the tests fail if you don't have database running for the back-end to connect to. The app won't load successfully if it is unable to connect to the database. We need to figure out how to have database running in github actions or maybe mock the database during the pipeline. Regardless, I believe @sebhani is currently investigating integration tests for the back-end, he might have some ideas on how to approach this. My initial thought would be to try and use docker to run a database

We can potentially use the in-memory database for the tests instead

felixlapierre (approved):
LGTM - I don't fully understand how the build file works, but given that the checks seem to be working, i think we can merge.

For the future we'll have to get tests set up as well, and make sure sonarqube sees the test coverage too.

FGRCL (approved):

PR #88, opened 2020-10-12
[#23] Delete and update devices endpoint

mohhef:
# General info

-Delete devices, encoders, decoders endpoint
-Update devices endpoint
-Refactor exception classes

**Issue number**: 23

**Note to reviewers**:
-Update primary key is not allowed currently, could be allowed if decided later
-There is no reason to update an encoder or decoder serial number thus updates are limited to updating the device attributes itself(except serial number)


mohhef:
Why did the sonar checks didnt get fired?

PR #89, opened 2020-10-12
[#23] Delete and update devices

### LINKED ISSUE #23: Service: Endpoint for updating and deleting one sender/receiver

mohhef:
# General info
-Delete devices, encoders, decoders endpoint
-Update devices endpoint
-Refactor exception classes

Issue number: 23

Note to reviewers:
-Update primary key is not allowed currently, could be allowed if decided later
-There is no reason to update an encoder or decoder serial number thus updates are limited to updating the device attributes itself(except serial number)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,37 @@
+name: Build
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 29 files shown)


ButlerJAL (dismissed):
This PR looks functionally GTG. After conversing with @mohhef and @FGRCL about how we could and should refactor our code base we agreed to eventually take the time to reduce the overall complexity of the code, and apply better coding practices. For example, eliminate low value external dependencies, and not mix SQL and java code by having SQL queries in a resource file.

ButlerJAL (approved):

sharon-ho (approved):
I don't have much to add besides that it looks like it does the job. I agree with Joshua that there should be refactoring done to reduce the complexity.

PR #103, opened 2020-10-20
Add Bug Template

sharon-ho:
# General info

**Issue number**: N/A

**Task description**: 
- Added bug template to repo

# Testing

**Test coverage**:
N/A

# Additional notes

**Note to reviewers**:
N/A

**To do before merging**:
N/A

### DIFF
--- a/.github/bug_report_template.md
+++ b/.github/bug_report_template.md
@@ -17,8 +17,8 @@
 <!-- Table below is mostly for frontend related bugs, use if applicable --> 
 |  Actual  |
 |:---------------:|
-| <!--** Please highlight/box the parts which do not match the mockups **--> testing1|
+| <!--** Please highlight/box the parts which do not match the mockups **--> |
 
 | Expected |
 |:---------------:|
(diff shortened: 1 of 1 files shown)


mohhef (approved):
Good initiative for everyone to follow the same template

jun-loke (approved):
That's a template. Why didn't you use the header notation tho? It's it easier to just have 
`# Section Title` instead of `**Section Title**`?

sharon-ho:
> That's a template. Why didn't you use the header notation tho? It's it easier to just have
> `# Section Title` instead of `**Section Title**`?

@arejayelle I decided against it since bug reports tend to have the section titles as bolded instead of as headers, and we've been using this format casually for a while now (including during Minicap). Also because there are a lot of sections, the large font size and the fact that there are several successively seems a bit too overwhelming visually and seems to distract from the content itself.

This is what it would look like if section headers were used instead:
image

sharon-ho:
@felixlapierre Now that the bug report template is merged, could you possibly add it as a template in the repo's settings?

PR #110, opened 2020-10-29
[#98] Add device fields

### LINKED ISSUE #98: Add missing device fields
In order to stream video between devices, we need a few fields that aren't currently in our model for devices

- **IP Address**: If it's possible to determine this on the server side rather than asking the device to provide it, we should do that
- **Port**: For the current sprint we can assume that the network is pre-configured, so we'll probably ask the device to specify its port now, and possibly remove this later when we have hole punching / rendezvous mode
- **Time since device last communicated**: For determining the device status. Find a better name than this. I think it would be a better design to store this instead of the device status. It will be up to the frontend to calculate the difference between the current time and the time of last communication, and if it's over a certain time (say, 1min) then consider the device offline.

This issue should include updating the sample senders and receivers to match the new API (and update the frontend if it has been merged before this task is over)

mohhef:
# General info

- Added new fields in devices
- Added a separate table for different ports
- now devices can have multiple ports
- to know how to make requests with proper JSON format refer to postman collection on slack

**Issue number**: 98


ButlerJAL (commented):
@mohhef I'm done resolving the merge conflicts. Make sure that I did not alter any of your changes.

mohhef:
> @mohhef I'm done resolving the merge conflicts. Make sure that I did not alter any of your changes.

All seems good. Its ready to be reviewed.

src/main/java/org/beanpod/switchboard/controller/ChannelController.java
```diff
@@ -0,0 +1,4 @@
+package org.beanpod.switchboard.controller;
+
+public class ChannelController {
```
ButlerJAL: Why is this class empty?

mohhef: I removed it. We don't need it as of right now.

mohhef: Removed it. We dont need it for now.

src/main/resources/application.properties
```diff
@@ -5,4 +5,6 @@ logging.level.org.hibernate.type.descriptor.sql.BasicBinder=TRACE
 #TODO if applicable, edit datasource properties (CAUTION: do not push changes to version control)
 spring.datasource.url=jdbc:mysql://localhost:3306/switchboard?useSSL=false&serverTimezone=UTC
 spring.datasource.username=root
-spring.datasource.password=root
\ No newline at end of file
+spring.datasource.password=root
+#spring.h2.console.enabled=true
+#spring.datasource.url=jdbc:h2:mem:testdb
```
ButlerJAL: Can you remove the commented out properties?

ButlerJAL (changes requested):
Can you merge master into your branch for the latest code analysis changes?

src/main/resources/application.properties
mohhef: resolved

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
```diff
@@ -81,4 +80,16 @@ public ResponseEntity createDecoder(@RequestBody @Valid DecoderEntity decoderEnt
         }
         return ResponseEntity.ok("Decoder with serial number " + serialNumber + " Deleted");
     }
+
+    @PutMapping("/{serialNumber}")
+    public ResponseEntity<DecoderDTO> updateDecoder(@PathVariable String serialNumber,
+                                                    @RequestBody DecoderDTO decoderDTO){
```
FGRCL: I don't think we need both the serialNumber and the decoderDTO here, since the serial number shoud already be in the decoderDTO.

ButlerJAL:
@mohhef I'm documenting the stack overflow reference you found and shared on Slack about [filtering out Lombok boilerplate code from our coverage analysis](https://stackoverflow.com/questions/44584487/improve-lombok-data-code-coverage) in this discussion for informational purposes.

ButlerJAL (dismissed):

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
```diff
@@ -81,4 +80,16 @@ public ResponseEntity createDecoder(@RequestBody @Valid DecoderEntity decoderEnt
...
+        Optional<DecoderEntity> decoder = decoderService.findDecoder(serialNumber);
+        if (decoder.isEmpty()) {
+            throw new ExceptionType.DeviceNotFoundException(serialNumber);
+        }
+        decoder.get().getInputs().clear();
```
FGRCL: Why do you clear these channels before resaving it?

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
```diff
@@ -87,12 +88,9 @@ public ResponseEntity createDevice(@RequestBody @Valid DeviceEntity device) {
         if (!serialNumber.equals(device.getSerialNumber())) {
             throw new ExceptionType.DevicePrimaryKeyRestriction(serialNumber);
         }
-        int response = service.updateDevice(serialNumber, device);
+        DeviceEntity deviceEntity = service.save(deviceMapper.toDeviceEntity(device));
+        return new ResponseEntity<>(deviceMapper.toDeviceDTO(deviceEntity), HttpStatus.OK);
```
FGRCL: I think you can use `ResponseEntity.ok(deviceMapper.toDeviceDTO(deviceEntity))` if you prefer.

src/main/java/org/beanpod/switchboard/entity/ChannelEntity.java
```diff
@@ -0,0 +1,28 @@
...
+
+@Getter @Setter
+@AllArgsConstructor
+@NoArgsConstructor
+@Entity(name="Channel")
+@JsonIgnoreProperties({"hibernateLazyIntializer", "handler", "out", "in"})
+public class ChannelEntity {
+
+    @Id
+    Integer port;
```
FGRCL: I'm not sure about using ports ad IDs, port are definitely not unique amongst sender and receivers.

src/main/java/org/beanpod/switchboard/entity/ChannelEntity.java
```diff
@@ -0,0 +1,28 @@
...
+    String name;
```
FGRCL: This property doesn't have `Column` annotation.

src/main/java/org/beanpod/switchboard/entity/DecoderEntity.java
```diff
@@ -21,8 +23,24 @@
...
+    @ManyToMany(fetch = FetchType.LAZY,
+            cascade = {
+                    CascadeType.PERSIST,
+                    CascadeType.MERGE
+            })
+    @JoinTable(
+            name="decoder_channel",
+            joinColumns = @JoinColumn(name="serial_number"),
+            inverseJoinColumns = @JoinColumn(name="port")
+    )
```
FGRCL: Is it required to have this `JoinTable`

src/main/java/org/beanpod/switchboard/repository/EncoderRepository.java
```diff
@@ -10,12 +10,10 @@
 @Repository
 public interface EncoderRepository extends JpaRepository<EncoderEntity, String> {
 
+    List<EncoderEntity> findAll();
     EncoderEntity save(EncoderEntity decoder);
-
+    Long deleteEncoderEntityBySerialNumber(String serialNumber);
     Optional<EncoderEntity> findEncoderBySerialNumber(String serialNumber);
```
FGRCL, line 16: Why you do add these methods? I was able to use the JpaRepository implementation.

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
```diff
@@ -42,14 +49,19 @@
...
         //stubbing device and decoder objects
         device1 = DeviceFixture.getDevice1();
         device2 = DeviceFixture.getDevice2();
-        decoder1 = DecoderFixture.getDecoder1(device1);
-        decoder2 = DecoderFixture.getDecoder2(device2);
+
+        setOfChannels = ChannelFixture.getSetOfChannels();
+
+        decoder1 = DecoderFixture.getDecoder1(device1,setOfChannels);
+        decoder2 = DecoderFixture.getDecoder2(device2,setOfChannels);
```
FGRCL: I think our fixtures should access ChannelFixture on their own to get the set of channels. That way you don't have to build anything in the test code.

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
```diff
@@ -130,8 +142,34 @@ final void testDeleteDecoder(){
     //When a decoder is unavailable in the DB
     @Test
     final void testDeleteDecoderNotExisting(){
-        assertThrows(DeviceNotFoundException.class, () -> {
+        assertThrows(ExceptionType.DeviceNotFoundException.class, () -> {
             decoderController.deleteDecoder("Not Available decoder");
         }, "DeviceNotFoundException should have been thrown.");
     }
+
+
```
FGRCL, line 138: There is 2 emtpy lines here.

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
```diff
@@ -130,8 +142,34 @@ final void testDeleteDecoder(){
...
+    //Test exceptions when updating encoder
+    @Test
+    final void testUpdateDecoderExceptions(){
+        DecoderDTO decoderDTO1 = decoderMapper.toDecoderDTO(decoder1);
+
+        //When device is unavailable in the DB
+        assertThrows(ExceptionType.DeviceNotFoundException.class, () -> {
+            decoderController.updateDecoder("3", decoderDTO1);
+        }, "DeviceNotFoundException should have been thrown.");
```
FGRCL: I think this test might be passing for the wrong reasons. There should be a mock here that explicitly mocks the `decoderService` to return an empty list.

src/test/java/org/beanpod/switchboard/controller/DeviceControllerTest.java
```diff
@@ -126,23 +124,23 @@ final void testDeleteDevice(){
...
     final void testUpdateDevice(){
-        DeviceDTO deviceDTO1 = deviceMapper.toDeviceDTO(device1);
 
         when(deviceService.findDevice("1")).thenReturn(java.util.Optional.of(device1));
-        when(deviceService.updateDevice("1",deviceDTO1)).thenReturn(Integer.valueOf(1));
-
-        ResponseEntity<String> response = deviceController.updateDevice("1", deviceDTO1);
+        device1.setStatus("Stopped");
+        when(deviceService.save(device1)).thenReturn(device1);
+        DeviceDTO deviceDTO1 = deviceMapper.toDeviceDTO(device1);
```
FGRCL, line 131: I'd recommend adding a `DeviceDTO` fixture in `DeviceFixture` instead of mapping it.

src/test/java/org/beanpod/switchboard/controller/EncoderControllerTest.java
```diff
@@ -44,15 +47,19 @@
...
         device1 = DeviceFixture.getDevice1();
         device2 = DeviceFixture.getDevice2();
-        encoder1 = EncoderFixture.getEncoder1(device1);
-        encoder2 = EncoderFixture.getEncoder2(device2);
+
+        setOfChannels = ChannelFixture.getSetOfChannels();
+
+        encoder1 = EncoderFixture.getEncoder1(device1, setOfChannels);
+        encoder2 = EncoderFixture.getEncoder2(device2, setOfChannels);
         listOfEncoders = EncoderFixture.getListOfEncoder(encoder1, encoder2);
```
FGRCL: Those fixtures should be initialized in the BeforeEach to avoid leaks in between tests. For example when you clear the encoder outputs

src/main/java/org/beanpod/switchboard/entity/ChannelEntity.java
mohhef: Ports are unique amongst themselves and not accross encoders and decoders

src/main/java/org/beanpod/switchboard/entity/ChannelEntity.java
mohhef: That would mean that itll map to the same variable name in the db

src/main/java/org/beanpod/switchboard/entity/DecoderEntity.java
mohhef: A join table defines a manytomany relationship.
It provides the ability for encoders to have multiple ports and for each unique port to belong to multiple encoders.

Think of channels table as a  table that just says what are the posible set of ports and the join table is what makes a many to many relationship possible

src/main/java/org/beanpod/switchboard/repository/EncoderRepository.java
mohhef: What do you mean? These are custom methods(named queries) provided by JPA.

src/test/java/org/beanpod/switchboard/controller/DeviceControllerTest.java
mohhef: Mapping it increases test coverage

src/main/java/org/beanpod/switchboard/repository/EncoderRepository.java
mohhef: https://docs.spring.io/spring-data/jpa/docs/1.5.0.RELEASE/reference/html/jpa.repositories.html#jpa.query-methods.named-queries

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
mohhef: yeah i think  you are right, fixed!

src/main/java/org/beanpod/switchboard/entity/ChannelEntity.java
felixlapierre: Agree with Francois here - multiple encoders may have channels that use the same port, but they will have different names, so we can't have them "share" the same entity in the table.

src/main/java/org/beanpod/switchboard/entity/DecoderEntity.java
felixlapierre: I think we'll need to change the relationship. Many-to-many won't work because one channel can only belong to one device. We can't have a channel belong to multiple devices, otherwise they will have to share the same name and streams, which doesn't make sense.

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
mohhef: i wanted to make it explicit that:
9090 update to 9080 that updates the decoders port but doesnt update the port itself.

It could be removed, ill remove it

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
```diff
@@ -130,8 +142,34 @@ final void testDeleteDecoder(){
...
             decoderController.deleteDecoder("Not Available decoder");
         }, "DeviceNotFoundException should have been thrown.");
     }
+
+
+    //When a encoder is available in the DB
+    @Test
+    final void testUpdateDecoder(){
+
+        when(decoderService.findDecoder("1")).thenReturn(java.util.Optional.of(decoder1));
```
felixlapierre: can't we just import optional & get rid of the `java.util.`?

src/test/java/org/beanpod/switchboard/controller/DeviceControllerTest.java
felixlapierre: Maybe we should have separate tests that check just the mapper. but since it's a really simple mapper, that might be overkill

src/test/java/org/beanpod/switchboard/dao/DecoderDaoImplTest.java
```diff
@@ -0,0 +1,85 @@
...
+        DecoderEntity decoderEntity = decoderDaoImpl.save(decoder1);
+        assertEquals(decoderEntity,decoder1,"Returned decoder is not equal mocked");
+    }
+
+    @Test
+    final void testFindDecoder(){
+        when(decoderRepository.findDecoderBySerialNumber("1")).thenReturn(java.util.Optional.of(decoder1));
+        Optional<DecoderEntity> encoderEntity = decoderDaoImpl.findDecoder("1");
+        assertEquals(encoderEntity,java.util.Optional.of(decoder1),"Returned encoderdecoder equal mocked");
+    }
```
felixlapierre: I think you might have a typo here, it should be `decoderEntity` instead of `encoderEntity` and the message should be `"Returned decoder equal mocked"`

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
```diff
@@ -81,4 +80,15 @@ public ResponseEntity createDecoder(@RequestBody @Valid DecoderEntity decoderEnt
         }
         return ResponseEntity.ok("Decoder with serial number " + serialNumber + " Deleted");
     }
+
+    @PutMapping
+    public ResponseEntity<DecoderDTO> updateDecoder( @RequestBody DecoderDTO decoderDTO){
```
felixlapierre, line 84: Some strange behaviour i came across when manually testing this endpoint
- It's possible to update details about a device through updating its decoder. Do we want to keep that behaviour?
- It's possible to change the serial number of a device through this endpoint, without changing the serial number of the decoder (and without changing the serial number of any encoder that might be associated with the device, too). We probably want to update all the serial numbers to match, OR prevent changing serial numbers altogether.

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -80,4 +81,15 @@ public ResponseEntity createEncoder(@RequestBody @Valid EncoderEntity encoderEnt
         }
         return ResponseEntity.ok("Encoder with serial number " + serialNumber + " Deleted");
     }
+
+    @PutMapping
+    public ResponseEntity<EncoderDTO> updateEncoder(@RequestBody EncoderDTO encoderDTO){
```
felixlapierre, line 85: I thought the update endpoints for encoders and decoders were part of issue #23? But these endpoints were not included in PR #89. We should look out for that in our code reviews in the future.

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
```diff
@@ -75,24 +76,15 @@ public ResponseEntity createDevice(@RequestBody @Valid DeviceEntity device) {
...
-            throw new ExceptionType.DevicePrimaryKeyRestriction(serialNumber);
-        }
-        int response = service.updateDevice(serialNumber, device);
-
-        if (response != 1) {
-            throw new ExceptionType.DeviceNotUpdated(serialNumber);
+            throw new ExceptionType.DeviceNotFoundException(device.getSerialNumber());
         }
-        return ResponseEntity.ok("Device updated");
+        DeviceEntity deviceEntity = service.save(deviceMapper.toDeviceEntity(device));
```
felixlapierre, line 85: Why did we change to using `save` instead of `updateDevice`?

felixlapierre (changes requested):

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
mohhef: fixed

src/main/java/org/beanpod/switchboard/entity/DecoderEntity.java
mohhef: so if one device is outputting through 8080, cant another device output through 8080 aswell? @felixlapierre

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
mohhef: fixed

src/test/java/org/beanpod/switchboard/controller/DeviceControllerTest.java
mohhef: its a mapper that is auto generated from mapstruct, i dont know if its a good idea to test manually.

src/test/java/org/beanpod/switchboard/controller/EncoderControllerTest.java
mohhef: fixed

src/test/java/org/beanpod/switchboard/dao/DecoderDaoImplTest.java
mohhef: fixed

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
mohhef: yes, fixed

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
mohhef: update device was a custome query, save is provided by the jpa interface

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
mohhef: fixed

src/main/java/org/beanpod/switchboard/entity/ChannelEntity.java
mohhef: not sure if i understand what you mean here.
multiple encoders can still have channels that use the same port.
if name is a problem it can be made as part of the primary key or removed.

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
mohhef: fixed

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
mohhef: the reason is written in the description of that pr

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
mohhef: good catch, ill look into the fix

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
```diff
@@ -10,6 +10,7 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.hateoas.EntityModel;
 import org.springframework.hateoas.server.mvc.WebMvcLinkBuilder;
+import org.springframework.http.HttpStatus;
```
ButlerJAL: This import is unused

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
```diff
@@ -9,6 +9,7 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.hateoas.EntityModel;
 import org.springframework.hateoas.server.mvc.WebMvcLinkBuilder;
+import org.springframework.http.HttpStatus;
```
ButlerJAL: This import is unused

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -11,6 +11,7 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.hateoas.EntityModel;
 import org.springframework.hateoas.server.mvc.WebMvcLinkBuilder;
+import org.springframework.http.HttpStatus;
```
ButlerJAL: This import is unused

ButlerJAL (changes requested):

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
felixlapierre: We decided to create a bug for this once the PR is merged

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
felixlapierre: Sorry, I missed that when i read it, thanks :)

mohhef:
Added a change to make fransois PR easier to adjust. both entities have now references to each other. Now channel CRUD operations could be fully manageable from the channel entity without needing to set attributes from the encoder or decoder entity.

src/main/java/org/beanpod/switchboard/entity/ChannelEntity.java
```diff
@@ -0,0 +1,34 @@
...
+import javax.persistence.*;
+
+@Getter @Setter
+@AllArgsConstructor
+@NoArgsConstructor
+@Entity(name="Channel")
+@JsonIgnoreProperties({"hibernateLazyIntializer", "handler", "out", "in"})
+public class ChannelEntity {
+    @Id
+    String name;
```
felixlapierre: If we have `name` as the ID, does that mean no two channels will be able to have the same name, even if they're on different devices? if so, we can't use name. i feel like we might have to generate our own primary key.

felixlapierre (commented):
Only 1 more thing, after that I'm ready to approve it

felixlapierre (dismissed):

src/main/java/org/beanpod/switchboard/entity/ChannelEntity.java
mohhef: i think that is what @FGRCL told me too. @FGRCL already generates an id on his pull request, so we can use @FGRCL's id as unique key

src/main/java/org/beanpod/switchboard/entity/ChannelEntity.java
mohhef: the issue was clarified

src/main/java/org/beanpod/switchboard/entity/DecoderEntity.java
mohhef: the issue was clarified

ButlerJAL (approved):
All changes since my last review look good to me

felixlapierre (approved):
My comments were addressed, LGTM!!

src/test/java/org/beanpod/switchboard/controller/DeviceControllerTest.java
FGRCL: I'm a little late to the party,  but from my point of view the mappers are a library and coverage of those should be done during integration tests.

PR #90, opened 2020-10-15
[#85] Add unit tests for PR 62

sebhani:
Testing Decoder, Encoder and Device controllers.

**Issue number**: #85 

**Task description**: Test controllers for sender and receiver

 - Add DecoderControllerTest class that contains the corresponding unit tests
 - Add EncoderControllerTest class that contains the corresponding unit tests
 - Add DeviceControllerTest class that contains the corresponding unit tests

**Note to reviewers**:
Check if I have not missed any test cases. Also, run the tests to ensure that all currently available tests pass.

PR #111, opened 2020-10-30
Standard Directory Layout

ButlerJAL:
… No code was modified.

# General info
**Task description**: 
 
Files have been moved to better conform with maven's standard directory layout.

# Testing

**Test coverage**:
n/a

# Additional notes

**Note to reviewers**:
CAUTION: When pulling and checking out this branch in your IDE chances are that you'll be faced with errors because many files have been moved. It is best to clone, checkout StandardDirectoryLayout branch, and then import as a new project into IDE. With that said, contents of files have not been changed. As there are no code changes, the review is mainly to ensure that you can adapt your IDE's and working branches to the new layout.

WARNING: Working branches will be faced with merge conflicts once this is merged into master. I'm comfortable with merge conflicts, so, send me a message if you need help adapting the layout of a working branch.

**To do before merging**:
n/a


ButlerJAL:
This PR does not seek to improve code quality. SonarCloud Code Analysis may be ignored for this PR. The analysis is the same as on master due to there being no changes in code.

.github/workflows/build.yml
```diff
@@ -12,7 +12,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          ref: ${{ github.event.pull_request.head.sha }}
```
mohhef, line 15: Why is this removed

ButlerJAL: This line did not actually solve the issue we were facing with #90 . Arguably it made it worst. It covered up rather than fixed compile errors. The actual problem we had with #90 was that it was not synced with master causing there to be compile errors for the merge build check. In retrospect, the compile errors in the build step was the correct behaviour to have.

Such a line may cause us to merge incompatible code to the master branch.

mohhef:
Typically what I saw in terms of project structure, is what we had before. The maven java maven project is separated from the frontend.
Either separated from the root level of the repository or in totally 2 different repositories.

If someone have some resources on why we are changing to this new structure can you provide some links or so.

Here are several examples on  java maven projects with front-end(i can provide more sample projects if needed):
https://github.com/swathisprasad/spring-boot-angular6-maven-project/blob/master/tutorial-web/pom.xml
https://frakton.com/utilizing-maven-front-end-plugin-for-angular-spring-boot/
https://github.com/oktadeveloper/okta-spring-boot-react-crud-example

typically you would have 3 poms:
one for the root
one in the frontend directory which does the npm stuff for you
and one for the backend directory

initializing the root pom would install all the packages needed for npm and maven in the children pom
otherwise, you would just initialize the pom needed for your directory

ButlerJAL:
> Typically what I saw in terms of project structure, is what we had before. The maven java maven project is separated from the frontend.
> Either separated from the root level of the repository or in totally 2 different repositories.
> 
> If someone have some resources on why we are changing to this new structure can you provide some links or so.
> 
> Here are several examples on java maven projects with front-end(i can provide more sample projects if needed):
> https://github.com/swathisprasad/spring-boot-angular6-maven-project/blob/master/tutorial-web/pom.xml
> https://frakton.com/utilizing-maven-front-end-plugin-for-angular-spring-boot/
> 
> typically you would have 3 poms:
> one for the root
> one in the frontend directory which does the npm stuff for you
> and one for the backend directory
> 
> initializing the root pom would install all the packages needed for npm and maven in the children pom
> otherwise, you would just initialize the pom needed for your directory

@mohhef Here is the official Maven reference on [standard directory layout](http://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html)

mohhef:
> > Typically what I saw in terms of project structure, is what we had before. The maven java maven project is separated from the frontend.
> > Either separated from the root level of the repository or in totally 2 different repositories.
> > If someone have some resources on why we are changing to this new structure can you provide some links or so.
> > Here are several examples on java maven projects with front-end(i can provide more sample projects if needed):
> > https://github.com/swathisprasad/spring-boot-angular6-maven-project/blob/master/tutorial-web/pom.xml
> > https://frakton.com/utilizing-maven-front-end-plugin-for-angular-spring-boot/
> > typically you would have 3 poms:
> > one for the root
> > one in the frontend directory which does the npm stuff for you
> > and one for the backend directory
> > initializing the root pom would install all the packages needed for npm and maven in the children pom
> > otherwise, you would just initialize the pom needed for your directory
> 
> @mohhef Here is the official Maven reference on [standard directory layout](http://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html)
@MrJCipherButtles 
I am not sure if this is the way to go with our project. I saw this in small projects were the stuff in the webapp directory would be heavily coupled with maven containing simple html and css files. Not a full react project with a separate NPM package manager.

ButlerJAL:
@mohhef Here is a [video tutorial](https://www.youtube.com/watch?v=FOTLcs_P3DA) I just found that better explains the maven standard directory structure. The value I am trying to get across with this PR is the principle of **convention over configuration**.

mohhef:
@MrJCipherButtles 
I totally understand the value you are trying to add. But i think its not simply a maven project. It's a java(maven), react(npm) project.
The way i saw webapp directory being utilized is to put plain source files like html,css,js.
But when a full project like react is to be setup. They are typically separated from what i saw.
Probably someone else can give their view.

FGRCL (dismissed):
It seemed weird to me at first too, but I found [this spring boot project example](https://github.com/spring-petclinic/spring-petclinic-reactjs ) that used the same layout. If it simplifies config I'm all for it.

ButlerJAL:
I'd like to address the reluctance of change because there is no point of merging this if we are not all convinced.

Ensuring convention is not a limiting factor. Having a React project within the conventional webapp folder simply respects the maven standard; otherwise, absolutely nothing changes. What seems to be the blocker to move forward is the build/packaging process for Maven and NPM.

When npm builds/packages the react project we are left with many JavaScript files that makes up our frontend. Similarly, maven does this for the backend with a jar package.

When it comes down to packaging both together, there are plugins in maven that can automate the build of a react project by pointing to the webapp folder to then package the frontend and the backend into one resulting .jar package for easy deployment.

It's really important to note that this PR does not add, limit, nor change anything other than help the project fit a maven recommended conventional mold. It's like how 'x', 'y', 'z' are the conventional variables in math, but sure we can use 'a' if we want.

felixlapierre (approved):
We decided to leave the frontend folder in the root. 
We looked at other projects to see whether they put React projects in src/main/webapp or in their own folder in the root. It looks like they usually put it in a root folder, but we'll see how each option interacts with Sonarqube. Joshua will try putting it in webapp in a fork of the repository.

mohhef (approved):

felixlapierre:
Merging despite Sonarqube check failing since this PR does not have code changes and does not affect the quality

Issue #59, opened 2020-09-29
Endpoint for user to initate a video stream


felixlapierre:
Add/Find/Update/Delete

PR #112, opened 2020-10-31
[#59] Add CRUD endpoints for streams

FGRCL:
# General info
Adds the following endpoints to the back-end service

**Issue number**: #59 

**Task description**: 
- get Streams
- get StreambyId
- create Stream
- update Stream
- delete Stream
 
# Testing

**Test coverage**:
Unit test for the controller and service layer. Mostly happy paths.

# Additional notes

**Note to reviewers**:

**To do before merging**:


service/switchboardapp/src/main/java/com/switchboard/app/entity/StreamEntity.java
```diff
@@ -0,0 +1,28 @@
+package com.switchboard.app.entity;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.*;
+
+import javax.persistence.*;
+import java.util.Base64;
+
+@Entity(name = "Stream")
+@Data
```
mohhef: We should replace Data with Setter, Getter & EqualsAndHashCode. The problem with tostring is it is infinite. It will call this.getdecoder and decoder will call this.getstream and so on.

You will face this problem when debugging the code.

More insight: https://thorben-janssen.com/lombok-hibernate-how-to-avoid-common-pitfalls/

service/switchboardapp/src/main/java/com/switchboard/app/entity/mapper/StreamMapper.java
```diff
@@ -0,0 +1,16 @@
...
+import org.openapitools.model.StreamModel;
+
+
+@Mapper(componentModel = "spring")
+public interface StreamMapper {
+    StreamDTO toDto(StreamModel streamModel);
+    StreamEntity toEntity(StreamDTO streamDto);
+    StreamDTO toDto(StreamEntity streamEntity);
+    StreamModel toModel(StreamDTO streamDto);
+}
```
mohhef: i think mappers should deal with the entity and the dto.

1. streamdto->entity
2. streamdtos-> entity
3. streamentity ->dto

service/switchboardapp/src/main/java/com/switchboard/app/entity/StreamEntity.java
```diff
@@ -0,0 +1,28 @@
...
+    @Column(name = "id")
+    private long id;
+
+    @ManyToOne
+    @JoinColumn(name = "encoder_id", referencedColumnName = "serial_number")
+    private EncoderEntity encoder;
+
+    @ManyToOne
+    @JoinColumn(name = "decoder_id", referencedColumnName = "serial_number")
+    private DecoderEntity decoder;
```
mohhef: if you look at our ER diagram a stream is associated with an encoder and a channel.
That means that to initiate a stream you would need an encoderID, port.
In the ER diagram built it is a many to many relationship between encoder and channel.
Actually for the purpose of our application i dont think we need a new entity at all.

mohhef: I think all we need is crud operations for the channel entity and to initiate a stream, You would have a manytomany relation from channel to encoder (which we already have on master) but add an attribute called vid which is the id of the video being streamed

mohhef (changes requested):

FGRCL:
Github actions doesn't allow sharing secrets to fork pull request on public repositories. A new PR was made from a main repository branch with the same changes. see #122.

FGRCL:
GitHub actions doesn't allow sharing secrets to forks for public repositories. A new PR was made with the same changes, from a branch on the main repo to fix this issue. See #122

PR #114, opened 2020-11-01
[#24] Added unit tests for DeviceList page

### LINKED ISSUE #24: Device list page
"As a user, I want to see a list of devices in the system"

Add a page that displays the list of devices in the system.

The senders and receivers should be displayed separately.

Each entry should contain basic information for the device. Minimum: the display name of the device.

felixlapierre:
# General info

**Issue number**: #24 
**Task description**: 

 - Added unit tests for DeviceList, bringing total coverage of frontend up to 81.71%
 - Refactored DeviceList to take a DataSource argument in the props containing the functions to call for retrieving sender/receiver information. This just adds dependency injection to make it easier to test.
- Added some CSS classes that are just used in the tests, to make it easier to retrieve key elements.
 
# Testing

**Test coverage**:

image


frontend/src/devicelist/ActionMenu.js
```diff
@@ -47,8 +47,8 @@ export default class ActionMenu extends React.Component {
                     keepMounted
                     open={Boolean(this.state.anchorElement)}
                     onClose={this.handleClose}
-                    anchorOrigin={{horizontal: 'right'}}
-                    transformOrigin={{ horizontal: 'right' }}
+                    anchorOrigin={{horizontal: 'right', vertical: 'center'}}
+                    transformOrigin={{ horizontal: 'right', vertical: 'center'}}
```
gdzooks: was there a point to this change? on my end it doesn't display the popup below the action menu like in the mockups; is it used in the tests?

frontend/src/devicelist/__tests__/DeviceList.test.js
```diff
@@ -0,0 +1,125 @@
...
+    var sortBy = document.querySelector("#sortBySelect");
+    expect(sortBy).not.toBe(null);
+
+    var senderTab = document.querySelector("#vertical-tab-0 span");
+    expect(senderTab).not.toBe(null);
+    expect(senderTab.textContent).toBe("Senders")
+
+    var senderTab = document.querySelector("#vertical-tab-1 span");
+    expect(senderTab).not.toBe(null);
+    expect(senderTab.textContent).toBe("Receivers")
```
gdzooks: should probably rename this var to `receiverTab` for clarity?

gdzooks (changes requested):
only requesting one change in the tests! i understood what you did & how you did it without much trouble! no experience with react unit testing either but looks good

wondering if we want to try to add test(s) for our coded style classes? eg. a device is "offline" so the status indicator should have "offline" as text and the style class should be "red statusText"?

frontend/src/devicelist/ActionMenu.js
felixlapierre: I only added this to get rid of a warning that appears when `anchorOrigin` or `transformOrigin` do not have a `vertical` property. I may have put the wrong value though...

frontend/src/devicelist/__tests__/DeviceList.test.js
felixlapierre: Oops, copy paste error, will fix

frontend/src/devicelist/ActionMenu.js
felixlapierre: Changed vertical to `top`, returns to the old behaviour in 374bc07

frontend/src/devicelist/__tests__/DeviceList.test.js
felixlapierre: Renamed in 374bc07

gdzooks (approved):
thanks for addressing my comments so fast! nice job :tada:

ButlerJAL (approved):
The changes look very clean to me, and I confirm that I could run it locally. Great code coverage as well!

PR #115, opened 2020-11-01
[#93] Fixed Github actions crashes on the build step

### LINKED ISSUE #93: Github actions crashes on the build step
**OS:** Windows 10

**Steps to reproduce:**

1. Push code on the PR #91 or #90

**Expected result:**

The github action runner should pass on the build step.

**Actual result:**

Return of a package not found error by the github runner.

**Visual proof:**

ButlerJAL:
# General info

Fixed Github actions crashes on the build step with separate embedded database configuration for testing

**Issue number**: #93 

**Task description**: 

 - Updated src/main/resources/application.properties
 - Created src/test/resources/application.properties
 - Delete src/main/resources/data.sql (clean up)
 - Added H2 Database dependency
 - Updated .github/workflows/build.yml

# Testing

**Test coverage**:
n/a

# Additional notes

**Note to reviewers**:
I changed the production database name to 'switchboard' instead of 'testdb'. Once we get a production database running remotely, it would be wise not to include such properties in a public repo.

**To do before merging**:
n/a

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -2,6 +2,8 @@ spring.jpa.hibernate.ddl-auto=update
 spring.jpa.properties.hibernate.format_sql=true
 logging.level.org.hibernate.SQL=DEBUG
 logging.level.org.hibernate.type.descriptor.sql.BasicBinder=TRACE
+#TODO replace spring.datasource.url value with remote production database url
 spring.datasource.url=jdbc:mysql://localhost:3306/switchboard?useSSL=false&serverTimezone=UTC
-spring.datasource.username=root
-spring.datasource.password=root
+#TODO Add database credentials
+spring.datasource.username=
+spring.datasource.password=
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -2,8 +2,7 @@ spring.jpa.hibernate.ddl-auto=update
 spring.jpa.properties.hibernate.format_sql=true
 logging.level.org.hibernate.SQL=DEBUG
 logging.level.org.hibernate.type.descriptor.sql.BasicBinder=TRACE
-#TODO replace spring.datasource.url value with remote production database url
+#TODO if applicable, edit datasource properties
 spring.datasource.url=jdbc:mysql://localhost:3306/switchboard?useSSL=false&serverTimezone=UTC
-#TODO Add database credentials
-spring.datasource.username=
-spring.datasource.password=
+spring.datasource.username=root
+spring.datasource.password=root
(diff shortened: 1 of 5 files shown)


src/main/resources/application.properties
```diff
@@ -1,10 +1,7 @@
-spring.h2.console.enabled=true
 spring.jpa.hibernate.ddl-auto=update
 spring.jpa.properties.hibernate.format_sql=true
 logging.level.org.hibernate.SQL=DEBUG
 logging.level.org.hibernate.type.descriptor.sql.BasicBinder=TRACE
-spring.datasource.url=jdbc:mysql://localhost:3306/testdb?useSSL=false&serverTimezone=UTC
+spring.datasource.url=jdbc:mysql://localhost:3306/switchboard?useSSL=false&serverTimezone=UTC
```
felixlapierre: I feel like it would be ideal if the database name (and password) were easily configurable locally. anyone have ideas on how to do that? maybe a third `application.properties` that's gitignored so we can override there? or just IDE runtime configuration overrides?

felixlapierre (dismissed):
Looks good to me - super excited to finally have code coverage working on the backend again. My one comment is more of a discussion starter, not a suggested change.

src/main/resources/application.properties
ButlerJAL: Yes, we should all have some `.properties` file that is gitignored which includes credentials.

ButlerJAL: Actually, on second thought, we can simply use the same `applications.properties` with TODO comment <INSERT CREDENTIALS> placeholder for the username and password property values

mohhef (dismissed):
Great addition.

src/main/resources/application.properties
ButlerJAL: @felixlapierre see last commit

ButlerJAL: We could explore better options down the road, but for now I think this works

felixlapierre: The problem with using TODO is that when we fill in the value locally it will always appear as a change that is ready to be committed. So you have to be careful not to accidentally commit your local username & password which is very annoying.

felixlapierre: IMO we can leave it as default values for now, which will work for most people's setups, and come up with a better solution in a later PR.

ButlerJAL: @felixlapierre done

felixlapierre (approved):
re-approving since my previous review was made stale, still LGTM

mohhef (approved):
Re-approving after verifying commits pushed after my initial approval.

Issue #117, opened 2020-11-03
Back-end can't connect to database if the database hasn't manually been logged in before


FGRCL:
**Operating system:** 
Should happen on all OS, confirmed on Linux and Windows

**Steps to reproduce:**

1. Start a mysql database
2. Start the backend application
3. The application fails to start because it can't connect to the database

**Expected result:**
- Back-end properly starts

**Actual result:**
- Application crashes at boot

felixlapierre:
Not an issue anymore since we have a development profile which allows the DB connections when user hasn't logged into the db before

PR #118, opened 2020-11-03
[#117] Allow DB connections when user hasn't logged in to the database before

FGRCL:
# General info
This enables the back-end to connect to the database even though the DB hasn't been logged in to before.
see https://dev.mysql.com/doc/dev/connector-net/8.0/html/P_MySql_Data_MySqlClient_MySqlConnectionStringBuilder_AllowPublicKeyRetrieval.htm

**Issue number**: #117

**Task description**: 

 - enable allowPublicKeyRetrieval for database connections

# Testing

**Test coverage**:
N/A

# Additional notes

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -3,6 +3,7 @@ spring.jpa.properties.hibernate.format_sql=true
 logging.level.org.hibernate.SQL=DEBUG
 logging.level.org.hibernate.type.descriptor.sql.BasicBinder=TRACE
 #TODO if applicable, edit datasource properties (CAUTION: do not push changes to version control)
-spring.datasource.url=jdbc:mysql://localhost:3306/switchboard?useSSL=false&serverTimezone=UTC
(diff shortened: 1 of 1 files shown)


felixlapierre:
According to https://mysqlconnector.net/connection-options/ 
>  Note that AllowPublicKeyRetrieval=True could allow a malicious proxy to perform a MITM attack to get the plaintext password, so it is False by default and must be explicitly enabled.

In the interest of following best security practices, I suggest that if you want to enable public key retrieva, you override the database connection string in your runtime configuration. The way I do it is
```
spring-boot:run "-Dspring-boot.run.arguments=--spring.datasource.password=password --spring.datasource.url=jdbc:mysql://localhost:3306/switchboard?useSSL=false&serverTimezone=UTC&allowPublicKeyRetrieval=true"
```

felixlapierre:
Closing because, even though this is a problem that is confusing when setting up for the first time, we probably want to fix this with alternative Spring profiles rather than adding a security risk to the default profile.

PR #134, opened 2020-11-07
[#59] Add CRUD endpoints for streams

### LINKED ISSUE #59: Endpoint for user to initate a video stream

FGRCL:
# General info

Adds the following endpoints to the back-end service
**Issue number**: #59 

**Task description**: 
- get Streams
- get StreambyId
- create Stream
- update Stream
- delete Stream

# Testing

**Test coverage**:

Unit test for the controller and service layer. Mostly happy paths.

# Additional notes

**Note to reviewers**:

Please pay special attention to these names. I coded this when I was tired and it's very possible I made mistakes.
Encoder <-> Decoder
OutputChannel <-> InputChannel
Stream <-> Channel

**To do before merging**:


mohhef:
There are 303 code smells :/

mohhef:
I think there are structural changes in the database, before reviewing the code can you update the E/R so I can undertand what you are trying to acheive?

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
```diff
@@ -38,7 +42,10 @@
 
     @GetMapping
     public List<DecoderDTO> retrieveAllDecoders() {
-        return decoderMapper.toDecoderDTOs(decoderService.getDecoders());
+        //TODO this is necessary to avoid a stackoverflow since Decoders reference input channels which reference Decoders. I suggest adding CRUD endpoints to channels and not referencing the channels in the Encoder/Decoder at all.
+        List<DecoderEntity> decoderEntity = decoderService.getDecoders();
+        decoderEntity.forEach(this::removeDecoderReferences);
```
mohhef, line 44: what is the purpose of line 47, if i understand what you are trying to do, you can use 
@JsonIgnoreProperties({"hibernateLazyIntializer", "handler", "X"}), x is the attribute name that references the other entity. that will prevent the infinete refrences

mohhef (dismissed):

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
mohhef: Or as you mentioned we can do a unidirectional mapping from channel to decoder/encoder.

FGRCL: I wasn't able to get JsonIngoreProperties to work, I tried a whole bunch of different ways. I think it doesn't want to work because I have to ignore a nested object, for example ingoring `encoder` in `inputs` int `EncoderEntity`. Also we have different ignore requirements depending on how we retrieve those objects. When we get an `De/EncoderEntity` we want to ignore `channel.en/decoder` and when we retrieve just the channel we want to ignore `outputs`/`inputs`.

FGRCL:
> There are 303 code smells :/

I think a lot of those is because @MrJCipherButtles enabled linting on sonar this week, I think most of those don't actually lower the code quality score. I like seeing them in action though!

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
FGRCL: Yes, the unidirectional mapping would be nice, but that would need an entire new controller and I didn't have time to do that.

felixlapierre:
I checked out the code locally and tried most of the endpoints esp. the ones we need to finish the sprint, all is working well on my end. i'll drop a review in the morning

contracts/api-spec.yaml
```diff
@@ -107,6 +108,23 @@ paths:
         '200':
           description: Successful deletion of the encoder
 
+  /encoder/{uuid}/stream:
+    get:
+      operationId: getEncoderStreams
+      description: Retrieve the sending streams for the encoder
```
felixlapierre, line 114: To clarify, this endpoint isn't implemented in this PR right?

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
felixlapierre: We still need to resolve the reference to the decoder when getting the stream details, so a unidirectional mapping or a json ignore wouldn't work i think. Really, we want some mechanism to say "don't put an object more than once in a response". Francois' workaround is acceptable for the moment but let's create a bug after we merge this PR so we can fix it later.

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
```diff
@@ -3,9 +3,12 @@
...
 import org.beanpod.switchboard.dto.DeviceDTO;
+import org.beanpod.switchboard.dto.EncoderDTO;
 import org.beanpod.switchboard.dto.mapper.DeviceMapper;
 import org.beanpod.switchboard.entity.DeviceEntity;
+import org.beanpod.switchboard.entity.OutputChannelEntity;
 import org.beanpod.switchboard.exceptions.ExceptionType;
+import org.hibernate.result.Output;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.hateoas.EntityModel;
 import org.springframework.hateoas.server.mvc.WebMvcLinkBuilder;
```
felixlapierre: I don't see any diff in this class other than these added imports, are they still being used?

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -4,9 +4,11 @@
 import org.beanpod.switchboard.dao.DeviceDaoImpl;
 import org.beanpod.switchboard.dao.EncoderDaoImpl;
 import org.beanpod.switchboard.dto.EncoderDTO;
+import org.beanpod.switchboard.dto.mapper.CycleAvoidingMappingContext;
```
felixlapierre: Is this import still being used?

src/main/java/org/beanpod/switchboard/dao/DecoderDaoImpl.java
```diff
@@ -16,6 +20,10 @@
     DecoderRepository decoderRepository;
 
     public DecoderEntity save(DecoderEntity decoderEntity) {
+        Optional.of(decoderEntity)
+                .map(DecoderEntity::getInputs)
+                .orElse(Collections.emptySet())
+                .forEach(inputChannelEntity -> inputChannelEntity.setDecoder(decoderEntity));
```
felixlapierre, line 26: 👍

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
```diff
@@ -0,0 +1,70 @@
...
+        StreamDTO streamDto = StreamDTO.builder()
+                .inputChannel(inputChannelDTO)
+                .outputChannel(outputChannelDTO)
+                .build();
+
+        StreamEntity streamEntity = mapper.toEntity(streamDto);
+
+        if(streamRepository.existsDuplicate(createStreamRequest.getInputChannelId(), createStreamRequest.getOutputChannelId())){
+            throw new ExceptionType.StreamAlreadyExistsException(createStreamRequest.getInputChannelId(), createStreamRequest.getOutputChannelId());
+        }
```
felixlapierre, line 49: We will need to add a similar check for channels in a later PR, i was able to create multiple channels with the same name / same port

src/main/java/org/beanpod/switchboard/exceptions/ExceptionType.java
```diff
@@ -2,34 +2,48 @@
...
     }
 
+    public static class StreamAlreadyExistsException extends RuntimeException {
+        public final static String MESSAGE_TEMPLATE = "Stream already exists between input %s%s and output %s%s";
+        public StreamAlreadyExistsException(Long inputChannelId, Long outputChannelId){
+            super(String.format(MESSAGE_TEMPLATE, ID, inputChannelId, ID, outputChannelId));
+        }
+    }
+
+    public static class StreamDoesNotExistsException extends RuntimeException{
```
felixlapierre: Typo, should be `StreamDoesNotExistException`

src/main/java/org/beanpod/switchboard/dto/mapper/CycleAvoidingMappingContext.java
```diff
@@ -0,0 +1,22 @@
+package org.beanpod.switchboard.dto.mapper;
+
+import org.mapstruct.BeforeMapping;
+import org.mapstruct.MappingTarget;
+import org.mapstruct.TargetType;
+
+import java.util.IdentityHashMap;
+import java.util.Map;
+
+public class CycleAvoidingMappingContext {
```
felixlapierre: Explain what this class does?

src/main/java/org/beanpod/switchboard/dto/mapper/ChannelMapper.java
```diff
@@ -0,0 +1,8 @@
+package org.beanpod.switchboard.dto.mapper;
+
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+
+@Mapper
+public class ChannelMapper {
+}
```
felixlapierre: Empty class?

src/main/java/org/beanpod/switchboard/dto/EncoderDTO.java
```diff
@@ -15,14 +16,15 @@
...
     private String serialNumber;
 
     @Temporal(TemporalType.TIMESTAMP)
     @JsonFormat(pattern="yyyy-MM-dd HH:mm:ss")
     private Date lastCommunication;
 
-    private DeviceEntity device;
-    private Set<ChannelEntity> outputs;
+    private DeviceDTO device;
+    private Set<OutputChannelDTO> outputs;
```
felixlapierre, line 26: 👍

src/main/java/org/beanpod/switchboard/dto/mapper/EncoderMapper.java
```diff
@@ -1,13 +1,15 @@
...
+import org.beanpod.switchboard.dto.OutputChannelDTO;
 import org.beanpod.switchboard.entity.EncoderEntity;
+import org.mapstruct.Context;
 import org.mapstruct.Mapper;
 import org.mapstruct.factory.Mappers;
 
 import java.util.List;
 
-@Mapper(componentModel = "spring")
+@Mapper(componentModel = "spring", uses = {DeviceMapper.class, OutputChannelMapper.class})
```
felixlapierre, line 10: Why do we need DeviceMapper now but not before?

src/main/java/org/beanpod/switchboard/entity/InputChannelEntity.java
```diff
@@ -0,0 +1,30 @@
...
+@JsonIgnoreProperties({"hibernateLazyIntializer", "handler"})
+public class InputChannelEntity {
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    @Column(name = "id")
+    private long id;
+
+    @OneToOne(optional = false, cascade = CascadeType.ALL)
+    @JoinColumn(name = "channel_id")
+    private ChannelEntity channel;
```
felixlapierre, line 36: I get that we're reducing duplication by having `ChannelEntity` hold name & port and be separate from `InputChannelEntity` and `OutputChannelEntity`. I think it might be worth getting rid of ChannelEntity and just having those fields in both InputChannel and OutputChannel. But let's discuss that refacoring in a later PR since changing the contract this late in the sprint will cause problems.

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
```diff
@@ -22,11 +24,12 @@
 import org.springframework.web.context.request.RequestContextHolder;
 import org.springframework.web.context.request.ServletRequestAttributes;
 
-import java.text.ParseException;
+import java.beans.Encoder;
```
felixlapierre: Java beans encoder 🤔

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
```diff
@@ -48,23 +52,23 @@
...
     }
 
     @Test
     final void testRetrieveAllDecoders(){
         when(decoderService.getDecoders()).thenReturn(listOfDecoders);
+        when(decoderMapper.toDecoderDTOs(any())).thenReturn(DecoderFixture.getDecoderDtos());
 
         List<DecoderDTO> allDecoders = decoderController.retrieveAllDecoders();
-        List<DecoderDTO> listOfExpectDTODecoders = decoderMapper.toDecoderDTOs(listOfDecoders); //covert List<DecoderEntity> to List<DecoderDTO>
+        List<DecoderDTO> listOfExpectDTODecoders = DecoderFixture.getDecoderDtos();
```
felixlapierre, line 67: You appear to be adding a return value to the mock on `decoderMapper` but stopped using decoderMapper in the test, is it being used inside one of the test objects?

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
```diff
@@ -139,25 +144,28 @@ final void testDeleteDecoderNotExisting(){
...
-
+        DecoderDTO decoderDto = DecoderFixture.getDecoderDto();
         when(decoderService.findDecoder("1")).thenReturn(Optional.of(decoder1));
         decoder1.getInputs().clear();
         when(decoderService.save(decoder1)).thenReturn(decoder1);
-        DecoderDTO decoderDTO1 = decoderMapper.toDecoderDTO(decoder1);
-        ResponseEntity<DecoderDTO> response = decoderController.updateDecoder(decoderDTO1);
+        when(decoderMapper.toDecoderEntity(decoderDto)).thenReturn(decoder1);
+        when(decoderMapper.toDecoderDTO(any())).thenReturn(decoderDto);
+        //DecoderDTO decoderDTO1 = decoderMapper.toDecoderDTO(decoder1);
```
felixlapierre: Commented code

src/test/java/org/beanpod/switchboard/fixture/EncoderFixture.java
```diff
@@ -1,33 +1,46 @@
...
+    public static EncoderEntity getEncoder1(DeviceEntity device){
+        try {
+            return new EncoderEntity(SERIAL_NUMBER,
+                    simpleDateFormat.parse("2020-10-31 05:05:05"),
+                    device,
+                    setOfChannels);
+        } catch (ParseException e) {
+            e.printStackTrace();
+            throw new RuntimeException("Unable to parse the date in Encoder 1");
+        }
```
felixlapierre, line 31: I guess this is because ParseException is a checked exception? Cause it's kind of weird to have a try catch in a fixture since it should always work

felixlapierre (commented):
I have a lot of misc comments, some of which are just "we should create an issue to track this debt", otherwise no major changes. Great job @FGRCL !!

contracts/api-spec.yaml
FGRCL: You're right, it's not implemented. It was just my idea of how we could implement it.

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
FGRCL: No they're not used, but at least they're being flagged by sonarqube now! Cleaned up.

src/main/java/org/beanpod/switchboard/dto/mapper/CycleAvoidingMappingContext.java
FGRCL: That's a leftover of somethin I tried to avoid cycles in mapstrcut mappers. I took the code from [here](https://github.com/mapstruct/mapstruct-examples/tree/master/mapstruct-mapping-with-cycles). It's not being used now though so I removed it.

src/main/java/org/beanpod/switchboard/dto/mapper/ChannelMapper.java
FGRCL: Another failed bug fix attempt. Removed.

src/main/java/org/beanpod/switchboard/dto/mapper/EncoderMapper.java
FGRCL: Previously we didn't use DeviceMapper so when Mapstruct generated the mapper and used internal private method to map de Device of the Encoders. Because we don't have any special mapping for the Device it boils down to pretty much the same code. But in the future if we need to change Device Mapping the changes will propagate to the *coder. For example, we could use the mapper to "reduce" the device into their respective *coder DTO like I did for the Input/Output channel and `Channel`.

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
FGRCL: Oops

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
FGRCL: Yes, `DecoderController` uses `DecoderMapper`.

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
FGRCL: Oops

src/test/java/org/beanpod/switchboard/fixture/EncoderFixture.java
FGRCL: Yes, the `simpleDateFormat.parse` throws that checked exception. I decided to catch it here instead of propagating it all de way to tests. Anyways it boils down to pretty much the same result.

felixlapierre (dismissed):
Comments addressed & code tested locally, LGTM!

felixlapierre (approved):

mohhef (approved):
I had trouble understanding the requirement trying to be achieved by this p/r. After a discussion with @felixlapierre now everything is clear. E/R diagram is updated.

In the upcoming p/r: 
-we can get rid of the stream entity, it can be autogenerated by hibernate without explicitly having it. 
-We should also agree on either contract first or contract last because it creates a lot of inconsistencies between what we already had and this p/r (i prefer contract last). 
-Some of the code is implemented in the Dao, while in the already existing code, some logic is implemented in the controller. we should agree on which logic goes where.

felixlapierre:
We will merge despite the Sonarqube quality gate failing.
Francois has fixed 100 out of 300 code smells, most of the remaining code smells were already on the Master branch before this PR was made. We will create a PR in Sprint 4 to address all the code smells in Soanrqube.

PR #158, opened 2020-11-17
[#141] Unable to delete device

### LINKED ISSUE #141: Unable to delete devices/encoders/decoders
**Delete Device**
Steps to reproduce:
- Register a device in the system
- Register a decoder or encoder associated with that device
- Delete the device

Expected result:
- The device is deleted

Actual result:
- Error: `Cannot delete or update a parent row: a foreign key constraint fails (`switchboard`.`input_channel`, CONSTRAINT `FK3x922bnyo03dhfg60iecycq26` FOREIGN KEY (`decoder_serial`) REFERENCES `decoder` (`serial_number`))`

**Delete Encoder/Decoder**
Steps to reproduce:
- Register a device in the system
- Register a decoder associated with that device
- Delete the decoder

Expected result:
- The decoder is deleted (without deleting the device)

Actual result:
- Though the response to the delete request indicates that the decoder has been deleted, subsequently getting that decoder by ID or getting all decoders will show that the decoder has not been deleted and is still in the system

### LINKED ISSUE #162: Refactoring Technical dept
**Acceptance crieria**

- Use of lombok builder instead of constructors
- Use of constructor injections instead of field injections
- Remove all #TODO labelled comments
- Fix dept related to circular reference in all the classes
- Move the mapping logic to the dao layer
- Refactor relationships(i.e. relationship directions, bi/uni directional)

mohhef:
# General info

Fix bugs
Remove unused code
Add channel endpoints
Use of builder by Lombok instead of constructors(in tests)

**Issue number**: 141

**Task description**: 

- Fix delete encoder and decoder bug
- Add a controller for channels
- Remove all todos (use jsonbackreference instead)
- Move mapping logic to DAO layer
- Bidirectional relationship with output/input channels with no StackOverflow, depending on how we want it, it can be very easily changed to unidirectional

**Flow:** 
1. create a device
2. create a decoder
3. create a channel
4. use createOutputChannel endpoint to map a decoder to channel
5. depending on how the team decides, we have 3 choices:
   - when fetching the encoder/decoder entity, the details about the input/output channel is in the response (bidirectional)
   - when fetching the encoder/decoder entity, only show the related attributes and attributes related to input/ouput channels you would have to use get endpoints from channels (e.g. get outputchannels that belong to decoder "abc")
   - have both

**Note to reviewers** : 
- This pr does not address code smells that were on master
- The autogenerating of the stream entity using hibernate instead of explicitly creating an entity is not addressed in this pr


src/main/java/org/beanpod/switchboard/controller/ChannelController.java
```diff
@@ -0,0 +1,120 @@
...
+    @PostMapping
+    public ResponseEntity<ChannelDTO> createChannel(@RequestBody ChannelDTO channel) {
+        Optional<ChannelDTO> channelLookup = channelService.findChannel(channel.getId());
+        if (channelLookup.isPresent()) {
+            throw new ExceptionType.DeviceAlreadyExistsException(channel.getId().toString());
+        }
+        return ResponseEntity.ok(channelService.save(channel));
+    }
+
+    @PostMapping("/input/{id}/serial/{serial}")
```
FGRCL: I would suggest to change `serial` to `decoder`, from my understanding of REST it's more "Restful" to use your models as parameters. Also, from the a client's point of view, I think it would make sense to also have a `decoder/{id}/input/` and `decoder/{id}/input/{id}`. What do you think? we can tackle in other story.

src/main/java/org/beanpod/switchboard/controller/ChannelController.java
```diff
@@ -0,0 +1,120 @@
...
+        DecoderDTO encoderDTO = decoderService.findDecoder(serial).
+                orElseThrow(() -> new ExceptionType.DeviceNotFoundException(serial));
+        InputChannelDTO inputChannelDTO = InputChannelDTO.builder()
+                .channel(channelDTO)
+                .decoder(encoderDTO)
+                .build();
+        return ResponseEntity.ok(channelService.saveInputChannel(inputChannelDTO));
+    }
+
+    @PostMapping("/output/{id}/serial/{serial}")
```
FGRCL: Pretty much the same comment.

src/main/java/org/beanpod/switchboard/controller/ChannelController.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @PostMapping
+    public ResponseEntity<ChannelDTO> createChannel(@RequestBody ChannelDTO channel) {
+        Optional<ChannelDTO> channelLookup = channelService.findChannel(channel.getId());
+        if (channelLookup.isPresent()) {
+            throw new ExceptionType.DeviceAlreadyExistsException(channel.getId().toString());
+        }
+        return ResponseEntity.ok(channelService.save(channel));
+    }
```
FGRCL, line 50: So, do we need to create a Channel then create an InputChannel and OutputChannel? If there's is no use case for creating just a channel, I'd suggest we only have the endpoints to create an InputChannel and OutputChannel and let the service handle the creation of the Channel "behind the scenes".

felixlapierre:
@sebhani and @FGRCL will be reviewing this

src/main/java/org/beanpod/switchboard/controller/ChannelController.java
```diff
@@ -0,0 +1,120 @@
...
+        List<ChannelEntity> channelEntities = channelService.getChannels();
+        return channelMapper.toChannelDTOs(channelEntities);
+    }
+
+    @GetMapping("/{id}")
+    public ResponseEntity<ChannelDTO> retrieveChannel(@PathVariable Long id) {
+        return channelService.findChannel(id).
+                map(ResponseEntity::ok).
+                orElseThrow(() -> new ExceptionType.DeviceNotFoundException(id.toString()));
+    }
```
FGRCL, line 41: Is there a way to retrieve input channels and output channels?

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
```diff
@@ -39,44 +31,27 @@
...
-        // TODO this is necessary to avoid a stackoverflow since Decoders reference input channels which reference
-        // Decoders. I suggest adding CRUD endpoints to channels and not referencing the channels
-        // in the Encoder/Decoder at all.
-        decoder.ifPresent(this::removeDecoderReferences);
-        EntityModel<DecoderDTO> resource = EntityModel.of(decoderMapper.toDecoderDTO(decoder.get()));
-
-        WebMvcLinkBuilder linkTo = linkTo(methodOn(this.getClass()).retrieveAllDecoders());
-        resource.add(linkTo.withRel("all-decoders"));
-        return ResponseEntity.ok(resource);
+        return ResponseEntity.ok(decoder.get());
```
FGRCL: You can leverage `Optional`'s functional API in a lot of the Controllers and Services if you want. Here's an example for this particular method.
```
return decoder.findDecoder(serialNumber)
    .map(ResponseEntity::ok)
    .orElseThrow(() -> new ExceptionType.DeviceNotFoundException(serialNumber + "/Decoder"))
```
Also great job taking care of that code debt.

src/main/java/org/beanpod/switchboard/dto/mapper/EncoderMapper.java
```diff
@@ -2,20 +2,16 @@
...
 import org.beanpod.switchboard.dto.EncoderDTO;
 import org.beanpod.switchboard.entity.EncoderEntity;
+import org.mapstruct.InjectionStrategy;
 import org.mapstruct.Mapper;
 import org.mapstruct.factory.Mappers;
 
 import java.util.List;
 
-@Mapper(componentModel = "spring", uses = {DeviceMapper.class, OutputChannelMapper.class})
+@Mapper(componentModel = "spring", uses = {DeviceMapper.class}, injectionStrategy = InjectionStrategy.CONSTRUCTOR)
```
FGRCL: why are we changing the injection strategy?

src/main/java/org/beanpod/switchboard/exceptions/ExceptionType.java
```diff
@@ -21,27 +21,15 @@ public DeviceNotFoundException(String s) {
...
-
-    public static class DeviceNotUpdated extends RuntimeException {
-        public DeviceNotUpdated(String s) {
-            super(SERIAL_NUMBER + s );
-        }
-    }
-
     public static class StreamAlreadyExistsException extends RuntimeException {
-        public final static String MESSAGE_TEMPLATE = "Stream already exists between input %s%s and output %s%s";
+        public static final String MESSAGE_TEMPLATE = "Stream already exists between input %s%s and output %s%s";
```
FGRCL, line 25: Good catch.

src/main/java/org/beanpod/switchboard/exceptions/ExceptionType.java
```diff
@@ -21,27 +21,15 @@ public DeviceNotFoundException(String s) {
...
         public StreamAlreadyExistsException(Long inputChannelId, Long outputChannelId) {
             super(String.format(MESSAGE_TEMPLATE, ID, inputChannelId, ID, outputChannelId));
         }
     }
 
     public static class StreamDoesNotExistException extends RuntimeException {
-        public final static String MESSAGE_TEMPLATE = "Channel %s%s does not exist";
+        public static final String MESSAGE_TEMPLATE = "Channel %s%s does not exist";
```
FGRCL, line 32: Good catch.

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
```diff
@@ -39,30 +31,21 @@
...
+    public ResponseEntity<DeviceDTO> retrieveDevice(@PathVariable String serialNumber) {
+        Optional<DeviceDTO> device = service.findDevice(serialNumber);
         if (device.isEmpty()) {
             throw new ExceptionType.DeviceNotFoundException(serialNumber);
         }
-        EntityModel<DeviceDTO> resource = EntityModel.of(deviceMapper.toDeviceDTO(device.get()));
-        WebMvcLinkBuilder linkTo = linkTo(methodOn(this.getClass()).retrieveAllDevices());
-        resource.add(linkTo.withRel("all-devices"));
-        return ResponseEntity.ok(resource);
-
```
FGRCL, line 52: You decided to remove the `quick links`. I thought It was a good idea, if we want to readd them I'd like to talk with you to establish some rule of thumbs to decide where they should go an what they should contain.

src/main/java/org/beanpod/switchboard/dao/ChannelDaoImpl.java
```diff
@@ -1,59 +1,86 @@
...
+    @Autowired
+    ChannelMapper channelMapper;
+    @Autowired
+    InputChannelRepository inputChannelRepository;
+    @Autowired
+    InputChannelMapper inputChannelMapper;
+    @Autowired
+    OutputChannelRepository outputChannelRepository;
+    @Autowired
+    OutputChannelMapper outputChannelMapper;
```
FGRCL: Do you know what are the advantages and disadvantages of using `private final` vs `autowired`. I think they prefered `private final` where I iinterned at but I don't know why.

src/main/java/org/beanpod/switchboard/dao/ChannelDaoImpl.java
```diff
@@ -1,59 +1,86 @@
...
-        removeChannelReference(inputChannelEntity);
-        return inputChannelMapper.toDto(inputChannelEntity);
+
+    public List<ChannelEntity> getChannels(){
+        return channelRepository.findAll();
+    }
+
+    public Optional<ChannelDTO> findChannel(Long id){
+        return channelRepository.
+                findChannelEntitiesById(id).map(channel -> channelMapper.toChannelDTO(channel));
```
FGRCL: For this map method, you can pass channelMappers' method directly with
```
channelMapper::toChannelDto
```
instead of
```
channel -> channelMapper.toChannelDTO(channel)
```
In the latter we're creating a new lambda function that forward directly to `channelMapper.toChannelDTO`. In the top version we pass the `channelMapper.toChannelDTO` directly, which avoids creating that lambda.

src/main/java/org/beanpod/switchboard/dao/ChannelDaoImpl.java
```diff
@@ -1,59 +1,86 @@
...
         OutputChannelEntity outputChannelEntity = outputChannelRepository.getOne(id);
-        removeChannelReference(outputChannelEntity);
-        return outputChannelMapper.toDto(outputChannelEntity);
+        return outputChannelMapper.toOutputChannelDTO(outputChannelEntity);
     }
 
-    public void saveOutputChannel(OutputChannelDTO outputChannelDto) {
-        outputChannelRepository.save(outputChannelMapper.toEntity(outputChannelDto));
+    public Long deleteOutputChannelById(Long id){
+        return outputChannelRepository.deleteOutputChannelEntitiesById(id);
```
FGRCL, line 71: Do you think we should return the saved object to all Create and Update endpoints?

src/main/java/org/beanpod/switchboard/dao/DecoderDaoImpl.java
```diff
@@ -18,21 +16,21 @@
...
-        return decoderRepository.save(decoderEntity);
+    public List<DecoderEntity> getDecoders() {
+        return decoderRepository.findAll();
     }
 
-    public Optional<DecoderEntity> findDecoder(String serialNumber) {
-        return decoderRepository.findDecoderBySerialNumber(serialNumber);
+    public Optional<DecoderDTO> findDecoder(String serialNumber) {
+        return decoderRepository.
+                findDecoderBySerialNumber(serialNumber).map(decoder -> decoderMapper.toDecoderDTO(decoder));
```
FGRCL: Same comment about the lambda.
```
decoderMapper::toDecoderDTO
```

src/main/java/org/beanpod/switchboard/dao/DeviceDaoImpl.java
```diff
@@ -15,12 +17,17 @@
...
+    public DeviceDTO save(DeviceDTO device) {
+        return deviceMapper.toDeviceDTO(deviceRepository
+                .save(deviceMapper.toDeviceEntity(device)));
     }
 
-    public Optional<DeviceEntity> findDevice(String serialNumber) {
-        return deviceRepository.findDeviceBySerialNumber(serialNumber);
+    public Optional<DeviceDTO> findDevice(String serialNumber) {
+        return deviceRepository.
+                findDeviceBySerialNumber(serialNumber).map(device -> deviceMapper.toDeviceDTO(device));
```
FGRCL:
```
deviceMapper::toDeviceDTO
```

src/main/java/org/beanpod/switchboard/dto/DecoderDTO.java
```diff
@@ -16,12 +18,11 @@
...
+    @NotNull
     private String serialNumber;
-
     @Temporal(TemporalType.TIMESTAMP)
     @JsonFormat(pattern="yyyy-MM-dd HH:mm:ss")
     private Date lastCommunication;
-
     private DeviceDTO device;
-    private Set<InputChannelDTO> inputs;
+    private Set<InputChannelEntity> input;
```
FGRCL: I think DTOs should only depend on other DTOs. Otherwise it's a breach of our layered architecture.

src/main/java/org/beanpod/switchboard/dto/EncoderDTO.java
```diff
@@ -16,13 +18,12 @@
...
+    @NotNull
     private String serialNumber;
-
     @Temporal(TemporalType.TIMESTAMP)
     @JsonFormat(pattern="yyyy-MM-dd HH:mm:ss")
     private Date lastCommunication;
-
     private DeviceDTO device;
-    private Set<OutputChannelDTO> outputs;
+    private Set<OutputChannelEntity> output;
```
FGRCL: Same thing about DTOs depending on DTOs

src/main/java/org/beanpod/switchboard/dto/mapper/OutputChannelMapper.java
```diff
@@ -3,26 +3,14 @@
...
 import org.mapstruct.Mapper;
-import org.mapstruct.Mapping;
-import org.openapitools.model.OutputChannelModel;
+import org.mapstruct.factory.Mappers;
 
-import java.util.Collection;
+import java.util.List;
 
-@Mapper(componentModel = "spring", uses = {EncoderMapper.class})
+@Mapper(componentModel = "spring",  uses = {EncoderMapper.class, ChannelMapper.class})
```
FGRCL, line 10: Good stuff

src/main/java/org/beanpod/switchboard/entity/InputChannelEntity.java
```diff
@@ -1,43 +1,30 @@
...
+    @OneToOne(optional = false, cascade = {CascadeType.MERGE})
     @JoinColumn(name = "channel_id")
     private ChannelEntity channel;
 
-    @ManyToOne(fetch = FetchType.LAZY, cascade = {
-            CascadeType.ALL
-    })
+    @ManyToOne(optional = false, fetch= FetchType.LAZY, cascade = {CascadeType.MERGE})
     @JoinColumn(name = "decoder_serial")
+    @JsonBackReference
```
FGRCL, line 28: Is this where all the magic happens in solving the circular reference?

src/main/java/org/beanpod/switchboard/entity/InputChannelEntity.java
```diff
@@ -1,43 +1,30 @@
...
 @Builder
 @Entity(name = "InputChannel")
 @JsonIgnoreProperties({"hibernateLazyIntializer", "handler"})
 public class InputChannelEntity {
     @Id
+    @NotNull
     @GeneratedValue(strategy = GenerationType.AUTO)
-    @Column(name = "id")
-    private long id;
+    private Long id;
```
FGRCL, line 20: Should we use the primitive wrappers in the entities instead of the primitives directly?

src/test/java/org/beanpod/switchboard/controller/ChannelControllerTest.java
```diff
@@ -0,0 +1,161 @@
...
+   void testRetrieveAllChannels() {
+      when(channelService.getChannels()).thenReturn(channelEntityList);
+      when(channelMapper.toChannelDTOs(any())).thenReturn(channelDtoList);
+      List<ChannelDTO> channelDTOS = channelController.retrieveAllChannels();
+      assertEquals(channelDtoList, channelDTOS);
+   }
+
+   @Test
+   void testRetrieveChannel() {
+      when(channelService.findChannel(4569L)).thenReturn(Optional.of(channelDto));
```
FGRCL: The ChannelFixture.CHANNEL_ID constant is public, so you can access is directly in your tests.

src/test/java/org/beanpod/switchboard/controller/ChannelControllerTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+   @BeforeAll
+   static void ChannelFixture() throws ParseException {
+      channelEntityList = List.of(ChannelFixture.getChannelEntity());
+      channelDtoList = List.of(ChannelFixture.getChannelDto());
+      encoderDto = EncoderFixture.getEncoderDto();
+      decoderDto = DecoderFixture.getDecoderDto();
+      channelDto = ChannelFixture.getChannelDto();
+      inputChannelDto = ChannelFixture.getInputChannelDto();
+      outputChannelDto = ChannelFixture.getOutputChannelDto();
```
FGRCL, line 59: I would recommend moving these initialization to @BeforeEach, just in case a test modifies one of the test data object so it won't fail the other tests. It makes our tests "leaky".

src/test/java/org/beanpod/switchboard/controller/ChannelControllerTest.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.List;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.initMocks;
+
+class ChannelControllerTest {
```
FGRCL, line 29: Overall, great test coverage.

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
```diff
@@ -42,124 +42,111 @@
...
-        decoder1 = DecoderFixture.getDecoder1(device1);
-        decoder2 = DecoderFixture.getDecoder2(device2);
-        listOfDecoders = DecoderFixture.getListOfDecoders(decoder1, decoder2);
+    void setup() {
         MockitoAnnotations.initMocks(this); //to be able to initiate decoderController object
     }
 
     @Test
-    final void testRetrieveAllDecoders(){
+    final void testRetrieveAllDecoders() throws ParseException {
```
FGRCL: It doesn't seen to me like there's anything in this test that throws a `ParseException`, is it still required?

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
```diff
@@ -42,124 +42,111 @@
...
+    final void testUpdateDecoder() {
+        when(decoderService.findDecoder("1")).thenReturn(Optional.of(decoderDto));
+        when(decoderService.save(decoderDto)).thenReturn(decoderDto);
         ResponseEntity<DecoderDTO> response = decoderController.updateDecoder(decoderDto);
-
         assertEquals(200, response.getStatusCodeValue());
     }
 
     //Test exceptions when updating encoder
+    @SneakyThrows
```
FGRCL: Is this still required?

src/test/java/org/beanpod/switchboard/dao/DecoderDaoImplTest.java
```diff
@@ -52,31 +50,35 @@ void setup(){
...
-        assertIterableEquals(deviceEntities, listOfdecoders, "list of expected and given decoders are not equal");
+        assertIterableEquals(deviceEntities, listOfdecoders);
     }
 
     @Test
     final void testDeleteDecoder(){
         when(decoderRepository.deleteDecoderEntityBySerialNumber("1")).thenReturn((long) 1);
         Long response = decoderDaoImpl.deleteDecoder("1");
-        assertEquals(response, (long)1, "Deleting decoder should return 1");
+        assertEquals(response, (long) 1d);
```
FGRCL: I think you can replace `(long) 1d` with 1L

src/test/java/org/beanpod/switchboard/fixture/DeviceFixture.java
```diff
@@ -1,36 +1,42 @@
...
 import java.util.ArrayList;
 import java.util.List;
 
 public class DeviceFixture {
     public static final String SERIAL_NUMBER = "1";
+    public static final String SERIAL_NUMBER2 = "2";
+    public static final String IP_ADDRESS = "127";
     public static final String DISPLAY_NAME = "Device #1";
+    public static final String DISPLAY_NAME2 = "Device #2";
     public static final String STATUS = "Running";
```
FGRCL, line 17: Good stuff

src/test/java/org/beanpod/switchboard/fixture/EncoderFixture.java
```diff
@@ -1,67 +1,62 @@
...
-            e.printStackTrace();
-            throw new RuntimeException("Unable to parse the date in Encoder 2");
-        }
+    public static EncoderEntity getEncoderEntity2() throws ParseException {
+        return EncoderEntity.builder()
+                .serialNumber(SERIAL_NUMBER2)
+                .lastCommunication(simpleDateFormat.parse("2020-10-31 05:05:05"))
+                .device(DeviceFixture.getDevice2())
+                .output(null)
+                .build();
```
FGRCL: Could we use `@SneakyThrows` on these instead of adding a throws delcaration ?

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
```diff
@@ -42,124 +42,111 @@
...
-    static private DecoderEntity decoder1, decoder2;
+    static private DeviceEntity device;
+    static private DeviceDTO deviceDto;
+    static private DecoderEntity decoder;
+    static private DecoderDTO decoderDto;
+
     static private List<DecoderEntity> listOfDecoders;
 
+
+    @SneakyThrows
```
FGRCL: I think we can remove this `SneakyThrows` if we add the same annotation in the fixtures.

src/test/java/org/beanpod/switchboard/controller/EncoderControllerTest.java
```diff
@@ -46,122 +40,109 @@
...
-    static private DeviceEntity device1, device2;
-    static private EncoderEntity encoder1, encoder2;
+    static private DeviceEntity device;
+    static private DeviceDTO deviceDto;
+    static private EncoderEntity encoder;
+    static private EncoderDTO encoderDTO;
     static private List<EncoderEntity> listOfEncoders;
 
+    @BeforeAll
+    static void encoderFixture() throws ParseException {
```
FGRCL: Could we remove this `throws ParseException` by using `SneakyThrows` in the fixtures?

src/test/java/org/beanpod/switchboard/fixture/DecoderFixture.java
```diff
@@ -15,57 +13,51 @@
...
-            e.printStackTrace();
-            throw new RuntimeException("Unable to parse the date in Decoder 1");
-        }
+    public static DecoderEntity getDecoderEntity2() throws ParseException {
+        return DecoderEntity.builder()
+                .serialNumber(SERIAL_NUMBER2)
+                .lastCommunication(simpleDateFormat.parse("2020-10-31 05:05:05"))
+                .device(DeviceFixture.getDevice1())
+                .input(null)
+                .build();
```
FGRCL: Could we use `@SneakyThrows` on these instead?

FGRCL (commented):
I have a few action items, but overall great job with the code quality and props for handling the code debt I introduced.

src/main/java/org/beanpod/switchboard/controller/ChannelController.java
mohhef: i chanced serial to decoder, not sure if you mean we should accept the model itself as a parameter(i.e. decoder object).
it would decrease the payload to just use their serial number

src/main/java/org/beanpod/switchboard/controller/ChannelController.java
mohhef: fixed

src/main/java/org/beanpod/switchboard/controller/ChannelController.java
mohhef: Yes, i would agree. But i am not sure if we need it or not, would have to discuss it.

src/main/java/org/beanpod/switchboard/controller/ChannelController.java
mohhef: I did not add these endpoints as currently, you can see the input and output channel an encoder/decoder uses by retrieving the encoder/decoder.
Let me know if its necessary to add them

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
mohhef: Yes you are right, i did use abit of the optional api in some of the other classes. I did not want to change all the classes at once. Ill do it incrementally thoughout the pr's. 
I changed the method you quoted tho!

src/main/java/org/beanpod/switchboard/dto/mapper/EncoderMapper.java
mohhef: good catch, forgot to remove

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
mohhef: Yes, it could be very useful. I also had it as a proof of concept.
The reason i removed it is because i found that it is 3+ maturity level:https://blog.restcase.com/4-maturity-levels-of-rest-api-design/#:~:text=The%20Richardson%20REST%20Maturity%20Model,3%20in%20this%20maturity%20model.
Which could be overkill(i.e. more advanced), but if we establish rules for that it would be nice!!

src/main/java/org/beanpod/switchboard/dao/ChannelDaoImpl.java
mohhef: yes good point fixed!

src/main/java/org/beanpod/switchboard/dao/DecoderDaoImpl.java
mohhef: fixed

src/main/java/org/beanpod/switchboard/entity/InputChannelEntity.java
mohhef: yeah!

src/main/java/org/beanpod/switchboard/entity/InputChannelEntity.java
mohhef: "If you use primitives it will always hold a default value, in this case 0L for long, even if the value is not there in the database. And if you use the wrapper object it will be having a null value if the value is not in the database or the entity is not persisted yet."

src/test/java/org/beanpod/switchboard/controller/ChannelControllerTest.java
mohhef: good advice, fixed

src/test/java/org/beanpod/switchboard/dao/DecoderDaoImplTest.java
mohhef: fixed

src/main/java/org/beanpod/switchboard/dao/ChannelDaoImpl.java
mohhef: I think its more convinent for the frontend, i save/update the entity, i get it back, i use it directly

src/main/java/org/beanpod/switchboard/dao/DeviceDaoImpl.java
mohhef: fixed

src/test/java/org/beanpod/switchboard/controller/ChannelControllerTest.java
mohhef: refactored all code to use fixtures, fixed

src/main/java/org/beanpod/switchboard/dao/ChannelDaoImpl.java
mohhef: After reading this, https://dzone.com/articles/spring-di-patterns-the-good-the-bad-and-the-ugly
i believe you are right and we should avoid field injections as much as possible.
I am debation wheather i should change all the injections in this pr or in the upcoming one

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
mohhef: getDecoderDtos, it has a date field that throws parseexcetption

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
mohhef: if removed, i have to handle the exception or throw manually

mohhef: ah i understand your point, i sneaky threw it at a deeper level. fixed

src/test/java/org/beanpod/switchboard/fixture/EncoderFixture.java
mohhef: yes fixed

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
mohhef: yes, fixed

src/test/java/org/beanpod/switchboard/controller/EncoderControllerTest.java
mohhef: yes, fixed

src/test/java/org/beanpod/switchboard/fixture/DecoderFixture.java
mohhef: YES SIR, fixed

src/main/java/org/beanpod/switchboard/dto/DecoderDTO.java
mohhef: This is a problem that i faced while coding, for this particular case changing the entity to dto would crash my program. I will try debugging it but I am aware of this problem

mohhef: For reference i think the problem is related to this: https://stackoverflow.com/questions/36223752/prevent-cyclic-references-when-converting-with-mapstruct
I am currently trying to solve the problem

mohhef: fixed, long one. phew

mohhef:
**To address next pr:**
Guideline for the use of HATEOS
Refactoring stream entity to be autogenerated

src/main/java/org/beanpod/switchboard/controller/ChannelController.java
FGRCL: Oh yeah, the way I phrased that was confusing. Keep it as is, `.../decoder/{serial}` to signify the serial number of the decoder of...

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
FGRCL: fantastic!

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
FGRCL: Makes sense, we can keep it in the backlog as a "nice to have".

src/main/java/org/beanpod/switchboard/dao/ChannelDaoImpl.java
FGRCL: I would say to change only those back for now to keep the PR delta at a minimum. But in the future anytime we open a class we should take that opportunity to change those field injection annotation to constructor injection.

mohhef: decided to delegate to the upcoming pr

src/main/java/org/beanpod/switchboard/dao/ChannelDaoImpl.java
FGRCL: Alright, let's apply that as a general design rule for our API.

src/main/java/org/beanpod/switchboard/dto/DecoderDTO.java
FGRCL: Oh yes, I did face that issue also. You've probably figured it out already, but you can go look at the generated class by mapstruct to help you debug stuff like that.

src/main/java/org/beanpod/switchboard/entity/InputChannelEntity.java
FGRCL: Keeping that one as a mental note

src/main/java/org/beanpod/switchboard/entity/InputChannelEntity.java
FGRCL: Got it, I'm going to keep that in mind.

src/test/java/org/beanpod/switchboard/controller/ChannelControllerTest.java
```diff
@@ -0,0 +1,160 @@
...
+      when(channelService.saveOutputChannel(any())).thenReturn(outputChannelDto);
+      when(channelService.findChannel(ChannelFixture.CHANNEL_ID)).thenReturn(Optional.of(channelDto));
+      when(encoderService.findEncoder("1")).thenReturn(Optional.of(encoderDto));
+      ResponseEntity<OutputChannelDTO> outputChannel = channelController.createOutputChannel(ChannelFixture.CHANNEL_ID, "1");
+      assertEquals(outputChannelDto, outputChannel.getBody());
+   }
+
+   @Test
+   void deleteOutputChannel() {
+      when(channelService.deleteOutputChannelById(4569L)).thenReturn(1L);
```
FGRCL: I think this `4569L` slipped through the cracks.

src/main/java/org/beanpod/switchboard/dao/ChannelDaoImpl.java
mohhef: i thought it would break tests and stuff, it ended up going well. I changed all field injections in all classes to constructor injection

src/test/java/org/beanpod/switchboard/controller/ChannelControllerTest.java
mohhef: cracks stopped leaking,fixed

FGRCL (approved):
LGTM!

ButlerJAL (approved):
There are many changes in this PR. And, from what I have seen, these are all good changes in the right direction.

As discussed in our meeting, there are a few concerns regarding confusion with DTOs and Entities seeming to overlap on one another responsibilities. Also, our DAOs are apparently acting as combined data and service layers which doesn't seem ideal to me. As agreed, we shall tackle these architecture issues in future PRs when possible.

This PR has gone above and beyond its original mandate and is getting monstrous making it harder to easily approve changes in a review because PR focus gets lost. This is something I strongly suggest we avoid in the future.

Issue #148, opened 2020-11-10
Fix all sonarcloud bugs and vulnerabilities, fix 50% code smells


felixlapierre:
Acceptance critieria:
Sonarcloud indicates on master branch
- 0 bugs
- 0 vulnerabilities
- 220 or less code smells

PR #160, opened 2020-11-18
[#149, #150] Code Format Checks

### LINKED ISSUE #149: Code format check
Acceptance criteria:
- Whenever code is pushed that has formatting errors, a failure or warning should be thrown in GitHub

### LINKED ISSUE #150: Use ESLint linter in the frontend

ButlerJAL:
# General info

**Issue number**: #149, #150 

**Task description**: 
 - Google Java Format Check
     - Added workflow using Google Java Format in Github Actions
     - Formatted java code using google-java-format
     - Configured Checkstyle to use the google_checks.xml ruleset
 - ESLint Check
     - Added workflow using ESLint in Github Actions
     - Configured ESLint to generate an ESLint report for importation into SonarCloud
     - Updated the version of dependencies to fix vulnerability
     - Formatted code with Prettier and ESLint

On every push, a check will validate that all Java source files in our repo respect the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html). And, another check will validate that all JavaScript source files in our repo respect [ESLint Rules](https://eslint.org/docs/rules/) with the addition of [Prettier Options](https://prettier.io/docs/en/options.html)

# Testing

**Test coverage**: n/a

# Additional notes

**Note to reviewers**:

 - New version of react-scripts causes build to fail on ESLint errors when running `npm start`; hence, we changed default ESLint rules we currently do not respect to throw warnings instead of errors (To be fixed in another PR for issue #148).
 - The high number of files changed is due to formatting commits. You should spend your review effort looking at non-formatting commits.
 - We decided to not use GitHub action automatic commit option because the SonarCloud check would not trigger properly, and also because some contributors expressed a preference for passive checks over active checks.

**To do before merging**:

 - All contributors must install the [google-java-format](https://github.com/google/google-java-format) as a command-line program or a plugin to your IDE
 - All contributors must familiarize themselves with the `npm run lint:fix` script for Prettier formatting, and ESLint to auto fix detected errors. ESLint can only automatically fix a subset of errors; thus, certain errors still need to be fixed manually in order to solve all errors

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,43 @@
+name: Lint
+
+on: push
+
+jobs:
+  run-linters:
+    name: Google Java Format
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v2 # v2 minimum required
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: "11" # Recommended: latest versions of Google Java Format require JDK 11+
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
+
+      # ESLint must be in `package.json`
+      - name: Install Node.js dependencies
+        run: npm install
+        working-directory: ./frontend
+
+      - name: Run Google Java Format
+        uses: axel-op/googlejavaformat-action@v3
+        with:
+          skipCommit: "true"
+          args: "--skip-sorting-imports --dry-run --set-exit-if-changed"
+          githubToken: ${{ secrets.GITHUB_TOKEN }} # Recommended if you use MacOS
+
+      - name: Run ESLint
+        uses: wearerequired/lint-action@v1
+        with:
+          github_token: ${{ secrets.github_token }}
+          # Enable ESLint
+          eslint: true
+          eslint_dir: ./frontend
+          eslint_extensions: "js,jsx"
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -16,7 +16,8 @@
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "react-router-dom": "^5.2.0",
-    "react-scripts": "3.4.3"
+    "react-scripts": "4.0.0",
+    "typescript": "^2.8.0"
   },
   "scripts": {
     "start": "react-scripts start",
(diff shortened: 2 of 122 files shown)


FGRCL:
These are just formatting changes right, no code change?

ButlerJAL:
> These are just formatting changes right, no code change?

correct

felixlapierre:
Myself and @FGRCL will review

ButlerJAL:
Temporarily I set the ESLint check to pass on warnings. Once merged, I'll fix all warnings in another PR for issue #148, and set the ESLint check to fail on errors and warnings.

frontend/package.json
```diff
@@ -13,10 +13,11 @@
     "enzyme-adapter-react-16": "^1.15.5",
     "jest-sonar-reporter": "^1.3.0",
     "prop-types": "^15.7.2",
-    "react": "^16.13.1",
+    "react": "^16.14.0",
```
felixlapierre: did we have to increase the react minimum version for it to work?

frontend/package.json
```diff
@@ -13,10 +13,11 @@
...
     "react-dom": "^16.13.1",
     "react-router-dom": "^5.2.0",
-    "react-scripts": "3.4.3"
+    "react-scripts": "4.0.0",
+    "typescript": "^4.0.5"
```
felixlapierre: why did we have to add typescript?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,43 @@
...
+      # ESLint must be in `package.json`
+      - name: Install Node.js dependencies
+        run: npm install
+        working-directory: ./frontend
+
+      - name: Run Google Java Format
+        uses: axel-op/googlejavaformat-action@v3
+        with:
+          skipCommit: "true"
+          args: "--skip-sorting-imports --dry-run --set-exit-if-changed"
```
felixlapierre, line 33: I couldn't find what these flags do in their documentation, I assume it's so that it doesn't actually change the code? Or what?

felixlapierre (dismissed):
LGTM, small comments. Can you also add a bit in the README about installing the google formatter & how to use it?

frontend/package.json
ButlerJAL: We were getting a vulnerability warning and other peer dependency warnings. I forget the exact dependency, but one of them required 16.14.0 minimum. Actually, it could of been for another formatting plugin I was testing that didn't make the final cut. Mind you we are using JSX, and version [16.14.0](https://github.com/facebook/react/blob/master/CHANGELOG.md#16140-october-14-2020) did add better support. Does not hurt to keep this minimum version, but do you prefer a return to ^16.13.1?

frontend/package.json
ButlerJAL: It's a transitive peer dependency to the eslint-plugin because ESLint can lint typescript code. If you were to npm install without typescript, you'll see the warning that tsutils which is a dependency of eslint-plugin requires typescript as a peer dependency. To avoid warnings, I added the dependency, but in our case it is not a must. I just don't like warnings. Should I remove?

.github/workflows/lint.yml
ButlerJAL: Here is the [documentation](https://github.com/axel-op/googlejavaformat-action) on the flags. Correct, we run the format without committing the changes automatically. Instead, we get a failed check with the names of the files that need formatting.

Also regarding `--skip-sorting-imports`, the google java format IntelliJ plugin requires a small extra configuration step for import sorting and it's possible that different versions of the google java format plugin may not sort imports exactly the same. In other words, it's slightly unclear what import sorting the GitHub Action enforces versus that of other google java format plugin versions. Basically, import sorting is skipped to avoid potential head scratching moments caused by potentially different google java format plugin implementations regarding import sorting. However, the check will still check if there are unused imports to remove.

ButlerJAL:
> LGTM, small comments. Can you also add a bit in the README about installing the google formatter & how to use it?

I added a new section to our README for code styling. The [README on google-java-format](https://github.com/google/google-java-format#using-the-formatter) has easy to follow instructions about installing & how to use the google formatter, so, I simply referenced their README within our README.

frontend/package.json
felixlapierre: No you can keep that

frontend/package.json
felixlapierre: I'm fine with upgrading the version

felixlapierre (dismissed):
Re-approving

FGRCL (dismissed):
Very happy to have automated linting!

felixlapierre (approved):
Re-approving stale review, note the bugs labeled by sonarcloud were not created by this code, they were found by the added linters.

FGRCL (approved):
Re-approving

PR #163, opened 2020-11-23
Improved linter rules for frontend

### LINKED ISSUE #150: Use ESLint linter in the frontend

felixlapierre:
# General info

**Issue number**: Related to #150 
**Task description**: 

 - Ignored line ending characters in Prettier
- Configured a bunch of prettier and eslint rules to be warnings instead of errors - like this, we won't get the build blocked for minor stuff like spacing, but we will still see warnings as code smells in pull requests. Should lower the impact on dev flow while keeping the linting relatively easy.
 - Added Jest to eslint globals to remove false positives related to jest globals being undefined in the eyes of the linter.

# Testing

No code changes

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -37,7 +37,8 @@
   "eslintConfig": {
     "env": {
       "browser": true,
-      "es6": true
+      "es6": true,
+      "jest": true
     },
     "extends": [
       "plugin:react/recommended",
@@ -58,6 +59,8 @@
       "react"
     ],
     "rules": {
+      "prettier/prettier": "warn",
+      "spaced-comment": "warn",
       "class-methods-use-this": "warn",
       "import/named": "warn",
       "no-plusplus": "warn",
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -75,7 +79,31 @@
...
+      "react/no-render-return-value": "error",
+      "react/no-string-refs": "error",
+      "react/no-unescaped-entities": "warn",
+      "react/no-unknown-property": "warn",
+      "react/react-in-jsx-scope": "warn",
+      "react/require-render-return": "error",
+      "react/jsx-equals-spacing": "warn",
+      "react/jsx-tag-spacing": "warn",
+      "react/jsx-indent": "warn",
+      "react/jsx-indent-props": "warn"
```
mohhef, line 106: did you try this plugin: https://github.com/bfanger/eslint-plugin-only-warn,

found here: https://stackoverflow.com/questions/50705508/is-it-possible-to-show-warnings-instead-of-errors-on-all-of-eslint-rules

felixlapierre: When i read the readme it looked like it downgrades all errors to warnings. If you look at the list I left some of them as errors because stuff like `no-render-return-value` is worth breaking the compile over. I got the impression the plugin would not let us do that. But i'll try it and see if it lets us still set some as error if we want.

jun-loke (approved):
Thank you!! <3

mohhef (approved):

ButlerJAL (approved):
Good changes, I cherry picked these two commits into #166

PR #167, opened 2020-11-25
[#58] Update device status

### LINKED ISSUE #58: Update status of devices when they poll the server
**Backend**
Create endpoints 
- `/encoder/{id}/poll` which is the endpoint an encoder should use to "call home". This endpoint should return the list of all streams involving the given encoder. It should also update the "lastCommunication" field of the decoder to be the current time.
- `/decoder/{id}/poll` which is the endpoint a decoder should use to "call home". This endpoint should return the list of all streams involving the given decoder. It should also update the "lastCommunication" field of the decoder to be the current time.

The contract should specify that these endpoints should only be called by the device itself. Leave room for us to add more information to this response. We should also make sure that the lastCommunication field can't be set when creating a device - it should be initialized to null.

** Frontend**
When displaying device information, consider the device status to be "Online" if the last communication is within the last 5 minutes. We should make this "5 minutes" value configurable in a configuration file / environment variable, not hardcoded. The device status should be "Pending" if the value of lastCommunication is null, and "Offline" if the last communication was more than 5 minutes ago.

**Sample Apps**
Update the sample apps to call the poll endpoints instead of getting all streams when they are polling the server.

felixlapierre:
# General info

**Issue number**: #58 

Note that this PR contains the changes in #165 as well, you can ignore those when reviewing.

**Task description**: 

 - Added two endpoints to the backend: `/decoder/{serial}/streams` and `/encoder/{serial}/streams`. These endpoints update the time of last communication of the encoder/decoder to the current instant, and return a list with all the streams that involve that device.
 - In the frontend, when getting the device information, the status of each device is no longer hardcoded to `online`. If the device's last communication is within the last ten minutes, it sets it to online. If it has been more than 10 minutes, the device status is marked offline. If the value of `lastCommunication` is null, it is set to `pending`. This would happen if the device has not retrieved its streams yet.
- Also did a small fix in the frontend, since the names of the inputs/outputs properties of decoders/encoders changed. Also added a protection so the page doesn't crash if a device has no channels.
 
image

# Testing

**Test coverage**:
Tests coming soon to this PR!

**To do before merging**:
- [x] Add unit tests to backend
- [x] Add unit tests to frontend


src/main/java/org/beanpod/switchboard/controller/DecoderController.java
```diff
@@ -67,10 +73,17 @@
...
     }
-    return ResponseEntity.ok(decoderService.save(decoderDTO));
+    return ResponseEntity.ok(decoderDao.save(decoderDTO));
+  }
+
+  @GetMapping("/{serialNumber}/streams")
+  public ResponseEntity<List<StreamModel>> getDecoderStreams(@PathVariable String serialNumber) {
+    List<StreamDTO> streamDtos = decoderService.getDecoderStreams(serialNumber);
+    List<StreamModel> streamModelList = streamMapper.toModel(streamDtos);
+    return ResponseEntity.ok(streamModelList);
```
FGRCL: We should wrap this is a catch or use `Optional` to throw a general error if anything wrong happens.

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -69,10 +69,16 @@
...
       throw new ExceptionType.DeviceNotFoundException(encoderDTO.getSerialNumber());
     }
-    return ResponseEntity.ok(encoderService.save(encoderDTO));
+    return ResponseEntity.ok(encoderDao.save(encoderDTO));
+  }
+
+  @GetMapping("/{serialNumber}/streams")
+  public ResponseEntity<List<StreamModel>> getEncoderStreams(@PathVariable String serialNumber) {
+    List<StreamDTO> streamDtos = encoderService.getEncoderStreams(serialNumber);
+    return ResponseEntity.ok(streamMapper.toModel(streamDtos));
```
FGRCL: Would be good to wrap this too

src/main/java/org/beanpod/switchboard/dto/mapper/StreamMapper.java
```diff
@@ -15,5 +17,9 @@
 
   StreamDTO toDto(StreamEntity streamEntity);
 
+  List<StreamDTO> toDto(List<StreamEntity> streamEntityList);
```
FGRCL: Can rename this  to `toDtos` since it's a list?

src/main/java/org/beanpod/switchboard/dto/mapper/StreamMapper.java
```diff
@@ -15,5 +17,9 @@
...
+
   StreamModel toModel(StreamDTO streamDto);
+
+  List<StreamModel> toModel(List<StreamDTO> streamDtoList);
```
FGRCL: Can rename this  to `toModels` since it's a list?

src/main/java/org/beanpod/switchboard/repository/StreamRepository.java
```diff
@@ -19,4 +20,10 @@
   boolean existsDuplicate(
       @Param(value = "inputChannelId") long inputChannelId,
       @Param(value = "outputChannelId") long outputChannelId);
+
+  @Query("SELECT s FROM Stream s WHERE s.outputChannel.encoder.serialNumber = :encoderSerial")
+  List<StreamEntity> getEncoderStreams(@Param(value = "encoderSerial") String encoderSerial);
+
+  @Query("SELECT s FROM Stream s WHERE s.inputChannel.decoder.serialNumber = :decoderSerial")
```
FGRCL, line 26: Short and sweet, I like it.

src/main/java/org/beanpod/switchboard/service/DecoderService.java
```diff
@@ -0,0 +1,9 @@
+package org.beanpod.switchboard.service;
+
+import org.beanpod.switchboard.dto.StreamDTO;
+
+import java.util.List;
+
+public interface DecoderService {
+  List<StreamDTO> getDecoderStreams(String decoderSerialNumber);
+}
```
FGRCL: New layer, I like it!

src/main/java/org/beanpod/switchboard/service/DecoderServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+  private final DecoderDaoImpl decoderDao;
+  private final DateUtil dateUtil;
+
+  @Override
+  public List<StreamDTO> getDecoderStreams(String decoderSerialNumber) {
+    Optional<DecoderDTO> decoderOptional = decoderDao.findDecoder(decoderSerialNumber);
+    if (decoderOptional.isEmpty()) {
+      throw new ExceptionType.DeviceNotFoundException(decoderSerialNumber);
+    }
+    DecoderDTO decoder = decoderOptional.get();
```
FGRCL: I believe you can replace this with a single `orElseThrow()`

src/main/java/org/beanpod/switchboard/service/EncoderServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+  private final StreamDaoImpl streamDao;
+  private final DateUtil dateUtil;
+
+  @Override
+  public List<StreamDTO> getEncoderStreams(String encoderSerialNumber) {
+    Optional<EncoderDTO> encoderOptional = encoderDao.findEncoder(encoderSerialNumber);
+    if (encoderOptional.isEmpty()) {
+      throw new ExceptionType.DeviceNotFoundException(encoderSerialNumber);
+    }
+    EncoderDTO encoderDto = encoderOptional.get();
```
FGRCL: Same thing, can be replace by a `orElseThrow`.

src/main/java/org/beanpod/switchboard/util/DateUtil.java
```diff
@@ -0,0 +1,17 @@
+package org.beanpod.switchboard.util;
+
+import org.springframework.stereotype.Component;
+
+import java.time.Instant;
+import java.util.Date;
+
+/**
+ * Utility class that returns the date of the current time in UTC time zone Useful for injecting
+ * test values for current time in tests.
```
FGRCL: that comment says it's for tests but you use it in a service.

src/test/java/org/beanpod/switchboard/controller/EncoderControllerTest.java
```diff
@@ -142,12 +148,28 @@ final void testUpdateEncoder() {
...
         () -> {
           encoderController.updateEncoder(encoderDto);
         });
   }
+
+  @Test
+  final void testGetEncoderStreams() {
+    when(encoderService.getEncoderStreams(any(String.class)))
+        .thenReturn(List.of(StreamFixture.getStreamDto()));
+    when(streamMapper.toModel(anyList())).thenReturn(List.of(StreamFixture.getStreamModel()));
```
FGRCL: we should add a fixture method to return a list of streams

FGRCL: I like to use the explicit call values when mocking instead of `any` because it helps catch more errors when changes are made

FGRCL (changes requested):
Only reviewed the backend for now, I have several suggestions.

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
felixlapierre: The CustomRestExceptionHandler should catch any other errors, but i'll wrap it in Optional to handle any null return values

src/main/java/org/beanpod/switchboard/dto/mapper/StreamMapper.java
felixlapierre: Renamed it to `toDtoList`

src/main/java/org/beanpod/switchboard/dto/mapper/StreamMapper.java
felixlapierre: Renamed it to `toModelList`

src/main/java/org/beanpod/switchboard/service/DecoderService.java
felixlapierre: backend is like an onion
many layers
make you cry

src/main/java/org/beanpod/switchboard/util/DateUtil.java
felixlapierre: I removed that part of the comment. I put it because this class only exists so i can do dependency injection in tests, but it's probably more misleading than anything else.

src/test/java/org/beanpod/switchboard/controller/EncoderControllerTest.java
felixlapierre: The reason I did this instead of putting explicit call values is that if the value changes, the method will return null instead and the error message becomes less evident.

Like this, if you call with the wrong thing, then it will fail at the stage of the `verify` with a nice little diff telling you what properties don't match.

frontend/src/api/DeviceApi.js
```diff
@@ -4,28 +4,49 @@ import * as SampleData from "./SampleData";
...
             sender.serialNumber,
-            "Last Communication",
+            lastCommunication,
             sender.device.ipAddress,
             sender.device.displayName,
-            "Online",
+            getStatus(sender.lastCommunication),
             channels,
-            ["Sample sender"]
+            ["Additional Device details go here"]
```
ButlerJAL, line 49: Is it appropriate to hard code such an extras field when returning new DeviceInfo on a HTTP GET request?

frontend/src/api/DeviceApi.js
```diff
@@ -41,17 +62,29 @@ export function getReceivers(callback) {
...
             receiver.serialNumber,
-            "Last Communication",
+            lastCommunication,
             receiver.device.ipAddress,
             receiver.device.displayName,
-            "Online",
+            getStatus(receiver.lastCommunication),
             channels,
-            ["Sample Reciever"]
+            ["Additional Device details go here"]
```
ButlerJAL, line 87: Same as previous review comment

felixlapierre:
Note that I renamed one of the `__tests__` folders to `tests`. The reason I did this is that I added `DeviceFixture.js` for creating test data objects. The test runner was trying to run my fixture class as a file containing tests, because it tries to run every .js file in a `__tests__` folder.

The fix was to rename the folder to `tests` since then it only runs files that end in `.test.js`. Had to adjust pom.xml to inform the code coverage checker aswell.

src/main/java/org/beanpod/switchboard/repository/StreamRepository.java
ButlerJAL: We possibly want to investigate the JPA method name query features to avoid writing SQL within our Java code. This comment applies for all other instances of writing SQL queries within our Java code base.

ButlerJAL (dismissed):
I approve these changes. I left a few comments pointing to things that we might want to investigate in a future task. Overall, I like the better separation between our data and service layers on the code you worked on. I see a few fixes were done end to end to update device status. The test coverage on the new code is good. Although there are more issues to solve regarding us reworking our implementation, these changes are a step in the right direction.

ButlerJAL (approved):
reapprove

src/test/java/org/beanpod/switchboard/controller/EncoderControllerTest.java
FGRCL: That's fair

FGRCL (approved):
My comments were adressed

PR #168, opened 2020-11-25
[#25] Add Device Details Page

### LINKED ISSUE #25: Detailed device information page
When clicking on a device in the list, the user should see detailed information. This should contain every bit of information that is relevant to the user (basically everything other than the internal database ID).

jun-loke:
# General info

**Issue number**: Closes #25 

**Task description**: 

 - Create Device Details page
 
 
| Original   | Updated|
|N/A | |

# Testing

**Test coverage**:
Pending..
image

# Additional notes

**Note to reviewers**:
Since the mockup is uncertain, I didn't bother make it match the mockup

Page now has 2 buttons to edit and delete a device. These are not connected to any back end

To access any device details page, go to the Devices page and click on the name of a device (flow still pending, but this is what I decided to go with)

I am ignoring the `console.error` smells because I don't have another way of checking if errors have been thrown. by props validation

I am not testing index.js because the current architecture makes it very hard to test. This should be addressed in a refactoring PR at some later point 

There are some smells in some files I didn't touch. I will not be handling those in this PR..

**To do before merging**:
- [x] Hook up routing
- [x] Add Testing
- [x] Remove `/Test` routing


gdzooks (commented):
i'll review properly once merge conflicts are resolved but the architecture you've built for this page looks good! 
in the meantime i have a few comments at first glance :,) 1 in general & 2 in the code
the folders usually start with a lowercase so i think you should rename `DeviceDetailsPage` to `deviceDetailsPage`. (i had trouble with the search fields issue because for some reason my `general` folder was locally named `General` and it broke all my imports lol :sob:)

frontend/src/DeviceDetailsPage/DeviceDetailsConciseRow.jsx
```diff
@@ -0,0 +1,69 @@
...
+      <>
+        <TableRow>
+          <TableCell>
+            {DeviceDetailsConciseRow.getPropertyDisplayName(name)}
+          </TableCell>
+          {name !== "channels" ? (
+            <TableCell align="center">{value}</TableCell>
+          ) : (
+            <TableCell>
+              {DeviceDetailsConciseRow.createInnerTable(value)}
```
gdzooks: i'd like if we could continue to use `StatusIndicator` when it comes to displaying statuses, if possible :,3

frontend/src/DeviceDetailsPage/DeviceDetailsConciseRow.jsx
```diff
@@ -0,0 +1,69 @@
...
+          );
+        })}
+      </TableContainer>
+    );
+  }
+
+  render() {
+    const { name, value } = this.props;
+    return (
+      <>
```
gdzooks, line 51: i've never seen `<>` on its own, what does this do? :O

frontend/src/createStream/StreamingPage.jsx
```diff
@@ -4,6 +4,8 @@ import PropTypes from "prop-types";
 import StreamingTable from "./StreamingTable";
 import DynamicBreadcrumb from "../general/DynamicBreadcrumb";
 import StreamList from "../streamlist/StreamList";
+import DeviceInfo from "../model/DeviceInfo";
```
ButlerJAL: remove unused import

frontend/src/createStream/StreamingPage.jsx
```diff
@@ -4,6 +4,8 @@ import PropTypes from "prop-types";
...
+import StreamInfo from "../model/StreamInfo";
```
ButlerJAL: remove unused import

frontend/debug.log
```diff
@@ -0,0 +1,2 @@
+[1123/221214.290:ERROR:directory_reader_win.cc(43)] FindFirstFile: The system cannot find the path specified. (0x3)
+[1126/171039.206:ERROR:directory_reader_win.cc(43)] FindFirstFile: The system cannot find the path specified. (0x3)
```
ButlerJAL: Is it appropriate to version control a `debug.log` file? You possibly may want to add in `.gitignore`

frontend/src/DeviceDetailsPage/TabPanels/DeviceDetailsActivityPanel.jsx
```diff
@@ -0,0 +1,6 @@
+import { Container } from "@material-ui/core";
+import React from "react";
+
+export default function DeviceDetailsActivityPanel() {
+  return <Container>TODO: Design activity log</Container>;
+}
```
ButlerJAL, line 6: I'm not sure whether or not this code is Work in Progress? Is it appropriate to push WIP code in master? You may want to remove this from this pull request, and include it in a future PR once complete. Better to keep code that is WIP in one of our working branches rather than in our master branch.

frontend/src/DeviceDetailsPage/DeviceDetailsPage.jsx
```diff
@@ -0,0 +1,64 @@
...
+        <Grid item xs={6}>
+          <DeviceDetailsTabTable tabs={tabs} device={device} />
+        </Grid>
+      </Grid>
+    </Container>
+  );
+}
+DeviceDetailsPage.defaultProps = {
+  location: { state: { device: getSampleSender() } }
+};
```
ButlerJAL, line 56: Is there a particular reason why we have decided to include a defaultProps here? I ask because I haven't seen this done elsewhere.

frontend/src/api/SampleData.js
```diff
@@ -103,6 +103,18 @@ export function getSenders(callback) {
...
+    "1:10:111:999",
+    null,
+    "123:456",
+    "Sender 1",
+    "Online",
+    sampleOutputChannels,
+    extras
+  );
+}
+
```
ButlerJAL, line 117: If defaultProps in `DeviceDetailsPage.jsx` is not to be common practice, this getSampleSender() function could be removed.

ButlerJAL (commented):
I've read the note to reviewers section.

I agree that old smells not caused by this PR should not influence its review. However, I did notice two unused import smells caused by this PR that should be removed.

Regarding the console.error smells note, I see that this is caused by new tests that were written to test prop validation. The way prop validation is being tested is not very convincing because how can we guarantee that the console.error was not called for some reason other than an invalid prop? It seems to me that the way these tests are written to check that the console.error was called can lead to both false-positives and false-negatives. This is because the expected result of theses tests are not implementation independent meaning that I could go to the code and change your PropTypes in such a way that your tests will continue to pass when they should fail, or fail when they should pass making these tests no better than not having tests at all.

Unit testing prop validation should be more robust in that the component created is deeply inspected to have the value types that the developer expects. We should not count on what the implementation says is an error. We should count on what we expect irregardless of implementation.

I hope it's clear that the smells themselves for using console.error is not the problem I worry about, but that our reliance on console.error in testing is what we should avoid for far greater issues.

frontend/src/DeviceDetailsPage/DeviceDetailsConciseRow.jsx
jun-loke: It's linter's version of `<React.Fragment/>`

frontend/src/DeviceDetailsPage/DeviceDetailsPage.jsx
jun-loke: This is for testing the page alone, since I had to do a lot of strange things to have a dynamic URL for this page which don't like the regular props method.

frontend/src/DeviceDetailsPage/TabPanels/DeviceDetailsActivityPanel.jsx
jun-loke: It's not a defined page, but we know that we want this tab to exist. There is no mockup. It's here because the tab is indicated but the contents is undefined. I don't have a way to demo that the tabs work without it. And we will want this file to exist later.

frontend/src/api/SampleData.js
jun-loke: Yes, but again, testing

frontend/debug.log
jun-loke: Thought it was already git ignored. will remove

frontend/src/DeviceDetailsPage/DeviceDetailsConciseRow.jsx
gdzooks: !? ok, interesting choice for the linter xD
thanks for the explanation~

frontend/debug.log
jun-loke: 291c469

frontend/src/DeviceDetailsPage/DeviceDetailsConciseRow.jsx
jun-loke: f2f4c04

jun-loke:
> I've read the note to reviewers section.
> 
> I agree that old smells not caused by this PR should not influence its review. However, I did notice two unused import smells caused by this PR that should be removed.
> 
> Regarding the console.error smells note, I see that this is caused by new tests that were written to test prop validation. The way prop validation is being tested is not very convincing because how can we guarantee that the console.error was not called for some reason other than an invalid prop? It seems to me that the way these tests are written to check that the console.error was called can lead to both false-positives and false-negatives. This is because the expected result of theses tests are not implementation independent meaning that I could go to the code and change your PropTypes in such a way that your tests will continue to pass when they should fail, or fail when they should pass making these tests no better than not having tests at all.
> 
> Unit testing prop validation should be more robust in that the component created is deeply inspected to have the value types that the developer expects. We should not count on what the implementation says is an error. We should count on what we expect irregardless of implementation.
> 
> I hope it's clear that the smells themselves for using console.error is not the problem I worry about, but that our reliance on console.error in testing is what we should avoid for far greater issues.
@MrJCipherButtles 

this is why the console.error evaluation is as it is.

image

 (yes, it's from a different PR, but the point still stands) As you can see I matched the string as close as I could, and yet it's still saying it's not matched. I've looked into the various formatting schemes of errors but haven't found a solution yet. and trust me, I have looked
Unfortunately, the props validation only calls `console.error` and doesn't NOT create the components/sub components (I've checked for that as well). Personally, having a test that checks that we do have an error being thrown is better than not having any, even if it's not the most ironclad thing ever. 

There's an argument for just fully trusting that Proptypes can and will handle it, but I'd rather have some personal checks in the tests. Which is why I went through with this method, that trusts that proptypes is doing it's job. But is also checking that something is being done.

frontend/src/DeviceDetailsPage/DeviceDetailsPage.jsx
```diff
@@ -0,0 +1,64 @@
...
+import DynamicBreadcrumb from "../general/DynamicBreadcrumb";
+import DeviceDetailsTabTable from "./DeviceDetailsTabTable";
+import DeviceInfo from "../model/DeviceInfo";
+import { getSampleSender } from "../api/SampleData";
+
+export default function DeviceDetailsPage(props) {
+  const {
+    location: {
+      state: { device }
+    }
```
gdzooks, line 15: why is the state nested inside location? :O

frontend/src/DeviceDetailsPage/TabPanels/DeviceDetailsNotesPanel.jsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Container } from "@material-ui/core";
+
+export default function DeviceDetailsNotesPanel(props) {
+  const { extras } = props;
+  return <Container>{extras}</Container>;
```
gdzooks, line 7: could use some padding to read more easily but we've already said devicedetails page is gonna go through some redesigns so i'm not requesting a change xD

frontend/src/model/DeviceInfo.js
```diff
@@ -16,4 +16,8 @@ export default class DeviceInfo {
     this.channels = channels;
     this.extras = extras;
   }
+
+  static getProperties() {
+    return ["name", "serialNumber", "status", "ip", "channels"];
```
gdzooks, line 21: any reason we're only returning these particular properties? i see this gets called in `DeviceDetailsConciseTable.jsx` and but then `DeviceDetailsConciseRow.jsx` has a case for `lastCommunication` and the default for `extras`?
if it's about returning more concise info maybe a rename to `getConciseProperties` is in order?

frontend/src/DeviceDetailsPage/__tests__/DeviceDetailsTabTable.test.jsx
```diff
@@ -0,0 +1,141 @@
...
+
+        wrapper = Enzyme.shallow(
+          <DeviceDetailsTabTable tabs={tabs} device={device} />
+        );
+
+        expect(wrapper.find(Tab)).toHaveLength(5);
+        expect(wrapper.find(TabPanelH)).toHaveLength(5);
+      });
+    });
+
```
gdzooks, line 115: i assume this line is extra since it's the only blank line before an `it` i've noticed so far

frontend/src/general/TabPanelH.jsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Box } from "@material-ui/core";
+
+export default function TabPanelH(props) {
```
gdzooks: took a second to understand what the H was - do you think it's worth it to rename this to `HorizontalTabs` to match `VerticalTabs`? or maybe rename `VerticalTabs` so we have consistent naming for our tab files?
actually the way verticaltabs works is different from this file so maybe verticaltabs warrants a refactor issue in the future 🤔 

**tldr** maybe write out the `Horizontal` instead of just `H` since you're importing it as `TabPanel` regardless

gdzooks (changes requested):
your test coverage is 👌  wonderful!! and nice job getting this all together so fast!
i'll reiterate that i'd prefer if our folder names were more consistent - starting with lowercase at least, but honestly most of my comments are clarifications for myself or questions -- nothing that i think warrants preventing this from merging
there's 2 more comments of @MrJCipherButtles 's that are about unused imports that i'll request changes for but once those are taken care of i'm ready & willing to approve!

frontend/src/model/DeviceInfo.js
jun-loke: When I was discussing this with @felixlapierre we agreed to remove those from the options to display until we had a better idea of what's going on with the Device Details page, since so much is still undetermined. I'm hesitant to remove the options for the Row because nothing is locked in yet and I don't want to have to rewrite it later.

Personally, it doesn't make sense for last communication to be in the core details of the device. It will still need to be displayed, yes, but personally, I don't find that it should be with the name/IP/serial number. Might make more sense to slap it in with the Status, TBH 

I agree with the rename tho, change coming soon

frontend/src/general/TabPanelH.jsx
jun-loke: I think vertical tabs should be renamed, since it's not the tabs themselves but the corresponding panels

but yeah TabPanelH renamed in 6a1b1a7a

frontend/src/createStream/StreamingPage.jsx
jun-loke: 36e41309

frontend/src/createStream/StreamingPage.jsx
jun-loke: 36e41309

frontend/src/DeviceDetailsPage/TabPanels/DeviceDetailsNotesPanel.jsx
jun-loke: tell that to the linter

frontend/src/DeviceDetailsPage/DeviceDetailsPage.jsx
jun-loke: because that's how it's passed in the dynamic page routing. I don't make the rules \shrug
https://learnwithparam.com/blog/dynamic-pages-in-react-router/

frontend/src/model/DeviceInfo.js
jun-loke: 913cc52

frontend/src/DeviceDetailsPage/TabPanels/DeviceDetailsNotesPanel.jsx
gdzooks: lsdjfalsdk i meant more like adding `padding=3em` or something so the frontend text is easier to read

gdzooks (dismissed):
lgtm! thanks for answering my questions~ :tada:

frontend/src/DeviceDetailsPage/TabPanels/DeviceDetailsNotesPanel.jsx
jun-loke: This isn't a finalized page, it's just a placeholder ATM so I don't think it's worth putting in the effort to make it nice since we're just going to scrap this anyways

jun-loke:
Sorry, a bunch of issues came up from a bad merge 😭 . Everything SHOULD be fixed now. Please re-review 
@gdzooks @felixlapierre

gdzooks (approved):
ouff good ole merges!! :sob:
anyway looked over the new changes quickly and ran it on my end - everything looks good!

felixlapierre (approved):
Tested out on my end and works, works great with the backend running too, LGTM!

PR #182, opened 2020-11-27
[#145] Add a field in the stream model to suggest the streaming mode to use

### LINKED ISSUE #145: Use SRT rendezvous mode to allow connecting devices on different networks without port forwarding
SRT rendezvous mode allows two devices on different networks to communicate without needing port forwarding by using a third party to initiate the connection. We should use this mode for initiating streams between the sample encoder and sample decoder.

SRT rendezvous mode requires that the sender and the receiver be using the same local port. I talked to the PM and it turns out that devices can pretty much freely choose which port they want to use for their video channels - it's not as static as I previously thought when we modeled our system. So, if we need to, we can let one device choose which port to use, and ask the other device to match the port it's using to the other device. If we do this, we should use the sender port. That means we're allowed to ignore the port selected in the receiver's input channel in order to get SRT rendezvous mode to work.

Of course, we can't use the same port if the sample apps are running on the same computer, so we should probably detect this case and not use SRT rendezvous mode if the devices are on the same local network.

** Acceptance Criteria**
This issue was re-scoped to only include the backend part of the task: #194 contains the work necessary in the sample apps.
This issue will be accepted if, when the sample apps get their streams, they have all the information necessary to stream to a device on a different network.

FGRCL:
# General info

Handle SRT rendezvous mode in the backend.

**Issue number**: #145 
CLOSES #145 

**Task description**: 
 
 
image

# Testing

**Test coverage**:

Changes some unit tests

# Additional notes

**Note to reviewers**:

**To do before merging**:


contracts/api-spec.yaml
```diff
@@ -222,6 +275,10 @@ paths:
       responses:
         '200':
           description: Successful creation of the decoder
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/StreamModel'
```
felixlapierre: Description should be "Successful creation of the stream"

src/main/java/org/beanpod/switchboard/entity/StreamEntity.java
```diff
@@ -35,4 +36,8 @@
   @OneToOne(optional = false, fetch = FetchType.LAZY)
   @JoinColumn(name = "output_channel_id")
   private OutputChannelEntity outputChannel;
+
+  @Column(name = "mode")
+  @NotNull
+  private StreamModeEntity mode;
```
felixlapierre: Joshua raised some concerns about whether this actually persists in the database properly, we might need to do some extra stuff to handle enums.

src/main/java/org/beanpod/switchboard/dto/mapper/DeviceMapper.java
```diff
@@ -15,4 +17,9 @@
   DeviceEntity toDeviceEntity(DeviceDTO deviceDTO);
 
   DeviceModel toDeviceModel(DeviceDTO deviceDto);
+
+  List<DeviceModel> toDeviceModels(List<DeviceDTO> deviceDtos);
+
+  @Mapping(source = "publicIpAddress", target = "publicIpAddress")
+  DeviceDTO toDeviceDTO(CreateDeviceRequest createDeviceRequest, String publicIpAddress);
```
felixlapierre: that's a bit of a weird mapping annotation, does that really grab it from the 2nd parameter?

src/main/java/org/beanpod/switchboard/dto/mapper/DeviceMapper.java
```diff
@@ -15,4 +17,9 @@
   DeviceEntity toDeviceEntity(DeviceDTO deviceDTO);
 
   DeviceModel toDeviceModel(DeviceDTO deviceDto);
+
+  List<DeviceModel> toDeviceModels(List<DeviceDTO> deviceDtos);
```
felixlapierre: I've been doing `toDeviceModelList` to better differentiate the individual/group mapping methods, 1 character is easy to miss

src/main/java/org/beanpod/switchboard/dto/StreamModeDTO.java
```diff
@@ -0,0 +1,6 @@
+package org.beanpod.switchboard.dto;
+
+public enum StreamModeDTO {
+  SRT,
+  SRT_RENDEZVOUS
```
felixlapierre: can we maybe come up with a better name for the `SRT` enum value? it's a bit strange to me because SRT Rendezvous is still SRT. I wouldn't be too annoyed if we kept it like this though.

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
```diff
@@ -58,4 +71,25 @@ public void updateStream(StreamDTO streamDto) {
...
     streamRepository.save(streamEntity);
   }
+
+  private boolean onSamePrivateNetwork(
+      InputChannelDTO inputChannelDto, OutputChannelDTO outputChannelDto) {
+    return inputChannelDto
+        .getDecoder()
+        .getDevice()
+        .getPublicIpAddress()
+        .equals(outputChannelDto.getEncoder().getDevice().getPublicIpAddress());
```
felixlapierre: The formatting on this one is really weird, maybe it would end up clearer to extract each into variables then compare.

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
```diff
@@ -31,20 +34,30 @@ public StreamDTO getStreamById(Long id) {
...
+        channelService.getOutputChannelById(createStreamRequest.getOutputChannelId());
+    StreamDTO.StreamDTOBuilder streamDtoBuilder =
+        StreamDTO.builder().inputChannel(inputChannelDto).outputChannel(outputChannelDto);
+
+    if (onSamePrivateNetwork(inputChannelDto, outputChannelDto)
+        || onLocalNetwork(inputChannelDto, outputChannelDto)) {
+      streamDtoBuilder.mode(StreamModeDTO.SRT);
+    } else {
+      streamDtoBuilder.mode(StreamModeDTO.SRT_RENDEZVOUS);
+    }
```
felixlapierre: Should we extract this into its own `selectMode` method or is that overkill?

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
```diff
@@ -58,4 +71,25 @@ public void updateStream(StreamDTO streamDto) {
...
+    return deviceDto.getPublicIpAddress().equals(deviceDto.getPrivateIpAddress())
+        || deviceDto.getPublicIpAddress().equals(LOOPBACK_IP_V4)
+        || deviceDto.getPublicIpAddress().equals(LOOPBACK_IP_V6);
+  }
+
+  private boolean onLocalNetwork(
+      InputChannelDTO inputChannelDTO, OutputChannelDTO outputChannelDTO) {
+    return deviceOnLocalNetwork(inputChannelDTO.getDecoder().getDevice())
+        && deviceOnLocalNetwork(outputChannelDTO.getEncoder().getDevice());
+  }
```
felixlapierre: I think this would be a good place to put a comment or two explaining how we determine if devices are on local networks, and why that affects our decision to use SRT mode or not.

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
```diff
@@ -58,4 +71,25 @@ public void updateStream(StreamDTO streamDto) {
...
+        .getDecoder()
+        .getDevice()
+        .getPublicIpAddress()
+        .equals(outputChannelDto.getEncoder().getDevice().getPublicIpAddress());
+  }
+
+  private boolean deviceOnLocalNetwork(DeviceDTO deviceDto) {
+    return deviceDto.getPublicIpAddress().equals(deviceDto.getPrivateIpAddress())
+        || deviceDto.getPublicIpAddress().equals(LOOPBACK_IP_V4)
+        || deviceDto.getPublicIpAddress().equals(LOOPBACK_IP_V6);
```
felixlapierre: `DeviceDTO::publicIpAddress` is not actually the public IP address of the device, but the IP address of the device as seen from the server. If the device is on the same network as the server, this will not take on the value of the public IP, but the value of the local IP.

Granted, this only comes into play for local testing scenarios, so we can probably leave the name. But a comment explaining why it's possible for the public IP to be localhost could be useful I think. Also, good job thinking of the ipv6 case.

src/test/java/org/beanpod/switchboard/controller/StreamControllerTest.java
```diff
@@ -22,19 +22,19 @@
 class StreamControllerTest {
   private StreamController streamController;
 
-  @Mock private StreamDaoImpl channelService;
-  @Mock private StreamMapper channelMapper;
+  @Mock private StreamDaoImpl streamService;
+  @Mock private StreamMapper streamMapper;
```
felixlapierre: I am embarrassed that this got through code review :facepalm:

src/test/java/org/beanpod/switchboard/controller/StreamControllerTest.java
```diff
@@ -22,19 +22,19 @@
...
     MockitoAnnotations.initMocks(this);
-    streamController = new StreamController(channelService, channelMapper);
+    streamController = new StreamController(streamService, streamMapper);
   }
 
   @Test
   void testGetChannels() {
     // given
-    when(channelService.getStreams()).thenReturn(StreamFixture.getIdList());
+    when(streamService.getStreams()).thenReturn(StreamFixture.getIdList());
```
felixlapierre: All the test method names in this class are wrong too and reference channels instead of streams

contracts/api-spec.yaml
```diff
@@ -269,18 +326,36 @@ components:
...
     DeviceModel:
       description: A device thats either an  encoder or decoder
       properties:
         serialNumber:
           type: string
           example: 'BtmC8ckj'
-        ipAddress:
+        publicIpAddress:
+          type: string
+          example: '178.156.52.3'
```
felixlapierre: This is a breaking change in the contract, please update the frontend to match.

src/test/java/org/beanpod/switchboard/dao/StreamDaoImplTest.java
```diff
@@ -88,6 +91,121 @@ public void testCreateChannel() {
     verify(streamRepository).save(streamEntity);
   }
 
+  @Test
+  public void testCreateChannelAlreadyExists() {
+    // given
+    CreateStreamRequest createStreamRequest = StreamFixture.getCreateStreamRequest();
```
felixlapierre: Test should be named `createStreamAlreadyExists` i think, and the rest of the test names need adjusting too.

src/test/java/org/beanpod/switchboard/dao/StreamDaoImplTest.java
```diff
@@ -88,6 +91,121 @@ public void testCreateChannel() {
...
+    when(streamMapper.toEntity(any())).thenReturn(streamEntity);
+    when(streamRepository.existsDuplicate(
+            createStreamRequest.getInputChannelId(), createStreamRequest.getOutputChannelId()))
+        .thenReturn(false);
+
+    // when
+    streamService.createStream(createStreamRequest);
+
+    // then
+    verify(streamRepository).save(streamEntity);
```
felixlapierre: I don't see any differences between this test and the next three, it doesn't seem to be asserting or checking anything? I feel like it should be verifying if the streamEntity has the right SRT mode.

felixlapierre (changes requested):
Got a few suggestions / requests here and there.

src/main/java/org/beanpod/switchboard/entity/StreamEntity.java
ButlerJAL: Because this is either SRT rendezvous or not SRT rendezvous you might want to use a Boolean for the mode field instead

src/main/java/org/beanpod/switchboard/dto/StreamModeDTO.java
ButlerJAL: If you’re inclined to agree with my previous suggestion, I would get rid of the enum class and rename the mode field to be a Boolean with isRendezvousMode = { true, false }

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
ButlerJAL: I second this suggestion

src/main/java/org/beanpod/switchboard/entity/StreamEntity.java
FGRCL: I chose an Enum to allow for more felxibility if we want to add a streaming modes in the future.

src/main/java/org/beanpod/switchboard/dto/mapper/DeviceMapper.java
FGRCL: Actually it seems like that annotation isn't even required. Mapstrcut will grab the publicIpAddress from the second parameter by default.

src/main/java/org/beanpod/switchboard/entity/StreamEntity.java
ButlerJAL: That’s one way of looking at the problem. I’m not convinced that Enums is the way to go.

Right now I’m only viewing the problem as us needing to implement the SRT protocol for streaming, but if we needed to extend our service to have compatibility with other streaming protocols I would more rely on some form of inheritance that creates a stream of a certain type allowing us to achieve a desired behaviour based on the stream object run type. If you have one stream class for all possible types of streams, and each streaming protocol has their own unique characteristics, then extension will increase our need for if-else logic, and most likely result in a bloated stream class with fields going unused depending on the stream protocol.

I’m mostly just thinking aloud and I’ve yet to gather all my thoughts. I might need to review our design goals. But, what I’m saying is that for now it may be sufficient to just think about the SRT streaming protocol unless our design must support multiple streaming protocols?

And, if we need to support multiple streaming protocols down the road, then take the time to evaluate the design tradeoffs of using an Enum field vs other design options.

ButlerJAL: Not to force your hand out of enums this is a suggestion rather than a change request, but I just wanted to point out some issues you are likely to encounter if you go down this route. A Boolean on the other hand seems to fit better with what you are currently trying to accomplish, and is not as likely to cause you issues.

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
FGRCL: Would you rather have `publicIpAddress` name `requestIpAddress` because in reality it's the request that we grab from the HTTP request.

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
FGRCL: Would you rather have publicIpAddress name requestIpAddress because in reality it's the request that we grab from the HTTP request.

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
```diff
@@ -10,65 +11,85 @@
...
+    return Optional.of(deviceDto)
+        .map(service::save)
+        .map(deviceMapper::toDeviceModel)
+        .map(ResponseEntity::ok)
+        .orElseThrow(this::getUnknownException);
+  }
+
+  private RuntimeException getUnknownException() {
+    return new RuntimeException(UNKNOWN_ERROR_MESSAGE);
   }
```
mohhef: i think we can move this to the exception class

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
```diff
@@ -10,65 +11,85 @@
...
+    Optional<DeviceDTO> deviceLookup = service.findDevice(createDeviceRequest.getSerialNumber());
     if (deviceLookup.isPresent()) {
-      throw new ExceptionType.DeviceAlreadyExistsException(device.getSerialNumber());
+      throw new ExceptionType.DeviceAlreadyExistsException(createDeviceRequest.getSerialNumber());
     }
-    return ResponseEntity.ok(service.save(device));
+
+    return Optional.of(createDeviceRequest)
+        .map(createRequest -> service.createDevice(createRequest, request.getRemoteAddr()))
+        .map(deviceMapper::toDeviceModel)
```
mohhef, line 61: This is more of something to address in the future instead of addressing in this pr.
We map the entities to dao then we map it models. why not map it to models directly in the dao package

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
```diff
@@ -31,20 +34,30 @@ public StreamDTO getStreamById(Long id) {
...
+    OutputChannelDTO outputChannelDto =
+        channelService.getOutputChannelById(createStreamRequest.getOutputChannelId());
+    StreamDTO.StreamDTOBuilder streamDtoBuilder =
+        StreamDTO.builder().inputChannel(inputChannelDto).outputChannel(outputChannelDto);
+
+    if (onSamePrivateNetwork(inputChannelDto, outputChannelDto)
+        || onLocalNetwork(inputChannelDto, outputChannelDto)) {
+      streamDtoBuilder.mode(StreamModeDTO.SRT);
+    } else {
+      streamDtoBuilder.mode(StreamModeDTO.SRT_RENDEZVOUS);
```
mohhef: i feel like some of the logic here could be in service level instead of dao level

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
mohhef: i feel like these three methods seem like they dont belong in the dao, more in a service package or a utility package

src/test/java/org/beanpod/switchboard/dao/StreamDaoImplTest.java
```diff
@@ -88,6 +91,121 @@ public void testCreateChannel() {
...
+  }
+
+  @Test
+  public void testCreateChannelSameDeviceIpV6() {
+    String otherPublicIpAddress = StreamDaoImpl.LOOPBACK_IP_V6;
+    CreateStreamRequest createStreamRequest = StreamFixture.getCreateStreamRequest();
+    InputChannelDTO inputChannelDto = ChannelFixture.getInputChannelDto();
+    OutputChannelDTO outputChannelDto = ChannelFixture.getOutputChannelDto();
+    outputChannelDto.getEncoder().getDevice().setPublicIpAddress(otherPublicIpAddress);
+    StreamEntity streamEntity = StreamFixture.getStreamEntity();
```
mohhef: i think we can put these in before each?

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
```diff
@@ -10,65 +11,85 @@
...
+        .map(ResponseEntity::ok)
```
mohhef, line 63: This is more of something to address in the future instead of addressing in this pr.
We map the entities to dao then we map it models. why not map it to models directly in the dao package

mohhef (changes requested):

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
FGRCL: The google formatter makes some weird decision sometimes, I guess it encourages us to extract stuff to variables.

src/test/java/org/beanpod/switchboard/dao/StreamDaoImplTest.java
FGRCL: The difference is the public IP of the createStreamRequest, it increases the code coverage for the streaming mode selection. I would like to test the streaming mode, but since the returned object is an object returned by a mock, the streaming mode returned will not depend on the public IP address. An option would be to use argument captors, but I've read that can be considered a code smell. In my opinion, we should use integration tests to test that behaviour.

contracts/api-spec.yaml
FGRCL: Fixed

src/main/java/org/beanpod/switchboard/entity/StreamEntity.java
FGRCL: I ended up changing it to `boolean isRendezvous` mode as you suggested. For reference, the Enum would previously be saved as its integer value, which I think could cause some problems if the Enum order is changed. In the future if we need to persist the enum, we should annonate the entity parameter with `@Enumerated`. For example, we could use `@Enumerated(EnumType.STRING)` to save the enum as a string.

src/main/java/org/beanpod/switchboard/dto/mapper/DeviceMapper.java
FGRCL: I changed this mapping method to `toDeviceModelList` but we should add that to our list of refactor. I know there are a few mappers that use "s" to differentiate between a single object and a collection of object.

src/main/java/org/beanpod/switchboard/dto/StreamModeDTO.java
FGRCL: I changed it to `boolean isRendezvous`, see my comment above for more details.

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
FGRCL: Alright, it's extracted.

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
FGRCL: It's extracted in the `StreamServiceImpl` now.

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
FGRCL: I added comments on the two methods you requested. I also moved them to `NetworkingUtil` and for some reason I feel a lot more comfortable adding comments to util classes rather than controller/service/dao classes.

src/test/java/org/beanpod/switchboard/controller/StreamControllerTest.java
FGRCL: It's fixed

contracts/api-spec.yaml
FGRCL: I added a new column in the device list page for the IP address.

src/test/java/org/beanpod/switchboard/dao/StreamDaoImplTest.java
FGRCL: Done.

src/test/java/org/beanpod/switchboard/dao/StreamDaoImplTest.java
FGRCL: Since I moved that logic to `NetworkingUtil` I got rid of those tests since that logic is now tested in `NetworkingUtilTest`. Much easier to test, I should do that more often.

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
FGRCL: I introduced a `UnknownException` for these. Good suggestion.

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
FGRCL: Good call, I added a `StreamService`.

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
FGRCL: Great suggestion, I moved them to `NetworkingUtil`. Much, much easier to test that way also.

src/test/java/org/beanpod/switchboard/dao/StreamDaoImplTest.java
FGRCL: Yep, I moved them.

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
FGRCL: In my mind the Controller should convert the `model` to a `dto` then send it to the service layer. The "Create...Request" are sort of exceptions to these rules because they don't have DTOs.

frontend/src/devicelist/__tests__/DeviceList.test.jsx
```diff
@@ -72,8 +72,9 @@ test("Device list renders sender page by default", () => {
     expect(rowElements[1].textContent).toBe(sender.name);
     expect(rowElements[2].textContent).toBe(sender.serialNumber);
     expect(rowElements[3].firstChild.textContent).toBe(sender.status);
-    expect(rowElements[4].textContent).toBe(sender.ip);
-    expect(rowElements[5].querySelector("button span svg")).not.toBe(null);
+    expect(rowElements[4].textContent).toBe(sender.privateIp);
+    expect(rowElements[5].textContent).toBe(sender.publicIp)
```
ButlerJAL: ESLint missing ";" warning should be fixed

frontend/src/model/ConvertDataFormat.js
```diff
@@ -30,7 +32,8 @@ export function convertToServiceObject(deviceInfo) {
     lastCommunication: deviceInfo.lastCommunication,
     device: {
       serialNumber: deviceInfo.serialNumber,
-      ipAddress: deviceInfo.ip,
+      publicIpAddress: deviceInfo.publicIp,
+        privateIpAddress: deviceInfo.privateIp,
```
ButlerJAL: ESLint indentation warning should be fixed

ButlerJAL (changes requested):
I find it strange that the following two files show as empty file additions in your PR:
- frontend/src/devicelist/HeadCells.js
- frontend/src/devicelist/SingleTableRow.js

Was something done to these files?

Other than the ESLint warnings I highlighted in my review to fix, my general concerns have been addressed. You can re-request a review for my approval when you feel that you have addressed all contributor concerns.

src/main/java/org/beanpod/switchboard/entity/StreamEntity.java
felixlapierre: sounds good! to clarify the design goals at the moment, we only need to support SRT.

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
```diff
@@ -10,65 +11,81 @@
...
+        service
+            .findDevice(deviceModel.getSerialNumber())
+            .orElseThrow(
+                () -> new ExceptionType.DeviceNotFoundException(deviceModel.getSerialNumber()));
+
+    return Optional.of(deviceDto)
+        .map(service::save)
+        .map(deviceMapper::toDeviceModel)
+        .map(ResponseEntity::ok)
+        .orElseThrow(() -> new ExceptionType.UnknownException(CONTROLLER_NAME));
```
felixlapierre: Regression: This code for the update device endpoint is broken, you simply re-save what was retrieved in the database, and don't use the deviceModel parameter. So nothing is done. I think you meant to map the incoming deviceModel to a DTO then save it.

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
felixlapierre: I don't think that helps much, let's stick with publicIpAddress.

felixlapierre (dismissed):

ButlerJAL (dismissed):

mohhef (dismissed):
LGTM!!!!!

felixlapierre (approved):

ButlerJAL (approved):

Issue #147, opened 2020-11-10
Search fields in device list page


felixlapierre:
Right now our device list page has UI for searching the list based on different fields, but these don't do anything.
** Acceptance Criteria **
We should be able to search by
- name
- serial number
- ip address
- status

PR #183, opened 2020-11-27
[#147] Search fields in device list page and migration from table to data table

sharon-ho:
# General info

**Issue number**: #147 

**Task description**: 
 - Now using a data table library called `material-table` instead of making our own tables with `material-ui`
 - Searching now works for all fields and column filters
- Removed header underline and added more padding between the breadcrumbs and the page title
- Using search and sort by features built into the library, therefore the dummy search bar and sort by selector are now removed from the page
- Added a button to export table to CSV or PDF, which is a feature from the library that seems to add value despite not being in the original mockups
 - Styling/UI changes that come with using the `material-table` library
 
 
| Original   | Updated|
|Screenshot 2020-11-27 155120|image|

# Testing

**Test coverage**: 100%

# Additional notes

**Note to reviewers**: N/A

**To do before merging**:
- [x] Tests need to be updated


sharon-ho:
Closed due to build breaking after reverting a commit, see #189 for the new PR.

PR #188, opened 2020-11-28
[#120] Add system test

### LINKED ISSUE #120: Add system tests
Add the first system tests to the system
First, we should run the backend and frontend together, put some data in the database, and verify that the data displays in the device list

Once the "view streams" story is done, we should test that after we create a stream it appears in the list of streams.

If possible, we should also check if we can run & register the sample encoder and decoder in a system test, and check that they appear in the list of devices. If that works, we should try and create a test that runs the sample encoder and decoder, creates a stream between them, and check that the stream starts successfully.

** Acceptance Criteria **
As long as we have one system test, we can consider this task done, but we will then have to create a task for each of those system tests that was not done in this task.

sebhani:
**Issue number**: #120 

**Task description**: 

 - Set up Selenium for Windows
 - Create a test to check if a specific encoder is available or not
 - Create a class to handle post requests
 - Update the pom.xml to include the selenium dependency 

# Additional notes
The following code can be used later when testing the creation of encoder/decoder manually.
```
        //populate Name, Serial Name and IP Address fields
        driver.findElement(By.id("nameSearch")).sendKeys("Sender Test10");
        driver.findElement(By.id("serialSearch")).sendKeys("1:10:111:999");
        driver.findElement(By.id("ipSearch")).sendKeys("123:456");

        //select Online status
        {
            WebElement element = driver.findElement(By.id("statusSearch"));
            Actions builder = new Actions(driver);
            builder.moveToElement(element).clickAndHold().perform();
        }
        {
            driver.manage().timeouts().implicitlyWait(3, TimeUnit.SECONDS);
            WebElement element = driver.findElement(By.cssSelector("#menu- > div:nth-child(1)"));
            Actions builder = new Actions(driver);
            builder.moveToElement(element).release().perform();
        }
        driver.findElement(By.cssSelector("body")).click();
        driver.findElement(By.cssSelector(".MuiListItem-dense:nth-child(1)")).click();

        driver.findElement(By.id("DeviceListAddDevBtn")).click();

        //Check if the device was added
        driver.navigate().refresh();
```

**Note to reviewers**:
The test right now is creating a device with serial number 10. Then, it's relating it to an encoder.
If you need to clean and install maven, you have two options:
  1- Run the front-end (i.e. npm start), then clean & install maven - I haven't tried this option, but it should work
  2- Comment out the Selenium test first, then clean & install maven


pom.xml
```diff
@@ -111,6 +111,12 @@
             <artifactId>jackson-databind-nullable</artifactId>
             <version>${openapi-jackson-nullable-version}</version>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.seleniumhq.selenium/selenium-java -->
```
FGRCL: I think we can remove this comment

pom.xml
```diff
@@ -111,6 +111,12 @@
...
+        <dependency>
+            <groupId>org.seleniumhq.selenium</groupId>
+            <artifactId>selenium-java</artifactId>
+        </dependency>
```
FGRCL: Let's explicitly add the dependency version here

src/test/java/system_tests/TrialSeleniumTest.java
```diff
@@ -0,0 +1,67 @@
...
+        System.setProperty("webdriver.chrome.driver","src\\test\\java\\system_tests\\chromedriver.exe");
+        driver = new ChromeDriver();
+
+        //create a test device
+        HashMap<String, String> deviceParam = new HashMap();
+        deviceParam.put("serialNumber","10");
+        deviceParam.put("displayName","Sender Test10");
+        deviceParam.put("status","Running");
+        deviceParam.put("ipAddress","212.150.5.74");
+        postRequest("http://localhost:8080/device",deviceParam);
```
FGRCL: There must be a way to reuse our fixtures here instead of defining the object by hand

src/test/java/system_tests/TrialSeleniumTest.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.TimeUnit;
+
+public class TrialSeleniumTest {
+    private static WebDriver driver;
+
+    @BeforeAll
+    static void setUp(){
+        //TODO set it up for mac/linux: https://stackoverflow.com/questions/228477/how-do-i-programmatically-determine-operating-system-in-java
+        System.setProperty("webdriver.chrome.driver","src\\test\\java\\system_tests\\chromedriver.exe");
+        driver = new ChromeDriver();
```
FGRCL: I did a little reasearch on the WebDriver and it seems that Selenium is pushing for the use of `RemoteWebDriver`. In which you start a server to act as your driver instead of using the chromedriver. Then you can connect to that server here using `RemoteWebDriver`. I think that could solve our compatibility issues on running the system tests on different platforms. Here's the documentation I was able to find [https://www.selenium.dev/documentation/en/remote_webdriver/](https://www.selenium.dev/documentation/en/remote_webdriver/), [https://github.com/SeleniumHQ/selenium/wiki/RemoteWebDriver](https://github.com/SeleniumHQ/selenium/wiki/RemoteWebDriver), and [https://github.com/SeleniumHQ/selenium/wiki/ChromeDriver](https://github.com/SeleniumHQ/selenium/wiki/ChromeDriver)

src/test/java/system_tests/TrialSeleniumTest.java
```diff
@@ -0,0 +1,67 @@
...
+        List<WebElement> devicesRows = encodersTable.findElements(By.tagName("tr")); //find all tr elements inside found table
+
+        boolean assertValue = false;
+        for(WebElement element : devicesRows){
+           if(element.getText().contains("Sender Test10")) {
+               assertValue = true;
+               break;
+           }
+        }
+
```
FGRCL, line 65: I think you might be able to replace this with 
```
assertValue = devicesRows.stream().anyMatch(element -> 
  element.getText().contains("Sender Test10") 
);
```
if you prefer

src/test/java/system_tests/HttpHandler.java
```diff
@@ -0,0 +1,55 @@
...
+        try{
+            Integer.parseInt(value);
+            return true; //value is int
+        }catch (Exception e){
+            return false; //value isn't int
+        }
+     }
+
+     //converts a hashMap to a string
+     private static String convertHashMapToString(HashMap<String, String> hashMap){
```
FGRCL: Does this convert a hashmap to JSON? if so I think you can just use `new JSONObject(map);` instead

src/test/java/system_tests/TrialSeleniumTest.java
sebhani: Good idea. I believe we need to set up selenium grid (https://www.guru99.com/introduction-to-selenium-grid.html) which enables you to run selenium tests remotely. I suggest opening another issue for setting up the selenium grid.

src/test/java/system_tests/TrialSeleniumTest.java
sebhani: Good point. I updated it accordingly.

src/test/java/system_tests/TrialSeleniumTest.java
sebhani: I'll use it since it's neater.

src/test/java/system_tests/HttpHandler.java
sebhani: You're right! new JSONObject(map).toString() would return the same result.

src/test/java/system_tests/TrialSeleniumTest.java
```diff
@@ -0,0 +1,70 @@
...
+  static void setUp() {
+    // TODO set it up for mac/linux:
+    // https://stackoverflow.com/questions/228477/how-do-i-programmatically-determine-operating-system-in-java
+    System.setProperty(
+        "webdriver.chrome.driver", "src\\test\\java\\system_tests\\chromedriver.exe");
+    driver = new ChromeDriver();
+
+    // create a test device
+    deviceEntity = DeviceFixture.getDevice1();
+    HashMap<String, String> deviceParam = new HashMap();
```
felixlapierre: Small change to avoid a Java warning - do the same for the other spot with new HashMap
```suggestion
    HashMap<String, String> deviceParam = new HashMap<>();
```

src/test/java/system_tests/HttpHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+    Response response;
+    try {
+      response = client.newCall(request).execute();
+    } catch (IOException e) {
+      response = null;
+      System.err.println(e);
+    }
+
+    return response;
```
felixlapierre: In general i think it's better practice to throw an exception here rather than print to err and return null. But since this method is used in the setup method of the test class, that might not be the right move... what does everyone else think?

src/test/java/system_tests/HttpHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+  // checks if a value is an integer
+  private static boolean isInteger(String value) {
+    try {
+      Integer.parseInt(value);
+      return true; // value is int
+    } catch (Exception e) {
+      return false; // value isn't int
+    }
+  }
```
felixlapierre: My intelliJ tells me this method is unused, if that is the case we should remove it.

felixlapierre (commented):
I have minor suggestions, after that i'd be good to approve. Good job :D

For anyone else who wants to try out the tests, keep in mind you need to already have the backend and the frontend running when you run the selenium tests. Also, if a device with serial number 1 already exists in your database, it may cause the tests to fail if its info is not the same as the one used in the test.

I'm not a fan of having the chromedriver executable in our source code, if it's possible to get rid of that we should, but i'm okay with having it in temporarily.

src/test/java/system_tests/HttpHandler.java
ButlerJAL: I'm not sure what to do here either?

src/test/java/system_tests/HttpHandler.java
ButlerJAL: Done.

src/test/java/system_tests/TrialSeleniumTest.java
ButlerJAL: Done.

ButlerJAL:
Solved merge conflicts and updated PR for review.

felixlapierre (dismissed):

src/test/java/system_tests/HttpHandler.java
FGRCL: I think we should throw the exception. This method is used in the setup to essentially create what is the fixture for the test. We shouldn't run the tests if the fixture is missing or incomplete.

FGRCL (commented):
apart from the small exception discussion everything looks OK to me

src/test/java/system_tests/HttpHandler.java
ButlerJAL: Agreed. Done.

FGRCL (approved):
My comments were answered 👍

felixlapierre (approved):
Re-approving

PR #189, opened 2020-11-28
[#147] Search fields in device list page and migration from table to data table

### LINKED ISSUE #147: Search fields in device list page

sharon-ho:
# General info

**Issue number**: #147 

**Task description**: 
 - Now using a data table library called `material-table` instead of making our own tables with `material-ui`
 - Searching now works for all fields and column filters
- Removed header underline and added more padding between the breadcrumbs and the page title
- Now using search and sort by features built into the library, therefore the dummy search bar and sort by selector are now removed from the page
- Added a button to export table to CSV or PDF which is a feature from the library that seems to add value despite not being in the original mockups
 - Styling/UI changes that come with using the `material-table` library
 
 
| Original   | Updated|
|Screenshot 2020-11-27 155120|image|

# Testing

**Test coverage**:
image
image

# Additional notes

**Note to reviewers**: N/A

**To do before merging**:
- [x] Tests need to be updated


frontend/src/devicelist/ContentsTable.jsx
```diff
@@ -50,23 +48,31 @@ export default class ContentsTable extends React.Component {
...
+          borderColor="#f1f1f1"
+        >
           <VerticalTabs
             value={value}
             setValue={this.handleValueChange}
             classes={classes}
           />
-          <TabPanel value={value} index={0}>
-            <DevicesTable devices={senders} />
+          <TabPanel value={this.state.value} index={0}>
```
felixlapierre: Use the `value` const variable declared on line 47. Same for when you use `this.state.senders` or `this.state.receivers`. This will eliminate some code smells in this file.

frontend/src/devicelist/DevicesDataTable.jsx
```diff
@@ -0,0 +1,142 @@
...
+  SaveAlt
+} from "@material-ui/icons";
+
+import MaterialTable, { MTableToolbar } from "material-table";
+
+import ChannelDetailsTable from "./ChannelDetailsTable";
+import StatusIndicator from "../general/StatusIndicator";
+import ActionMenu from "./ActionMenu";
+
+export default class DevicesDataTable extends React.Component {
```
felixlapierre: I think we can keep the old name `DevicesTable` here. Or even `DeviceTable`.

frontend/src/devicelist/DevicesDataTable.jsx
```diff
@@ -0,0 +1,142 @@
...
+      toolbar: true,
+      search: true,
+      exportButton: true,
+      headerStyle: {
+        backgroundColor: "#f1f1f1",
+        fontWeight: "bold"
+        // position: 'sticky', top: 0
+      },
+      // maxBodyHeight: '650px',
+      // above lines commented out b/c scrollbar bug, see: https://github.com/mbrn/material-table/issues/780
```
felixlapierre: From what I can tell in the linked issue, these lines are CSS that allow us to have sticky headers. Why not uncomment them?

felixlapierre (commented):
Looking great! I only have a couple of suggestions.

Two more points I didn't find a good spot to put into comments:

- Please resolve the code smells highlighted by eslint. If you need help with that send me a message.
- I noticed very few changes to the tests were made. Is this because the old tests still validate the new table properly? I'm worried that even though the coverage is high, we aren't actually validating that stuff is working properly.

frontend/src/devicelist/DevicesDataTable.jsx
sharon-ho: There'll be a second scrollbar that makes it so that when that scrollbar is activated it no longer stays sticky. It's hard to explain but I could provide a GIF.

frontend/src/devicelist/DevicesDataTable.jsx
sharon-ho: Good point, I could do that. I initially named it differently because I didn't want to delete the old file/table just yet, but since that's been deleted I could rename it as `DeviceTable`.

frontend/src/devicelist/DevicesDataTable.jsx
felixlapierre: >     * I noticed very few changes to the tests were made. Is this because the old tests still validate the new table properly? I'm worried that even though the coverage is high, we aren't actually validating that stuff is working properly.

Nevermind this point, I somehow missed the devicelist test file when reviewing >.<

felixlapierre:
>     * I noticed very few changes to the tests were made. Is this because the old tests still validate the new table properly? I'm worried that even though the coverage is high, we aren't actually validating that stuff is working properly.

Nevermind this point, I somehow missed the devicelist test file when reviewing >.<

felixlapierre (dismissed):
LGTM!

felixlapierre (approved):
Re-approving

jun-loke (approved):
Code looks fine and couldn't break it on my end, Good job!

Just one question though, why did you opt to keep the scrolling when there's an option for pagination?

sharon-ho:
> Code looks fine and couldn't break it on my end, Good job!
> 
> Just one question though, why did you opt to keep the scrolling when there's an option for pagination?

@arejayelle For no reason in particular actually since we weren't using pagination before, it can be added though if you think it would be a better addition. I'll probably make a PR if we decide to add it in the future.

jun-loke:
> > Code looks fine and couldn't break it on my end, Good job!
> > Just one question though, why did you opt to keep the scrolling when there's an option for pagination?
> 
> @arejayelle For no reason in particular actually since we weren't using pagination before, it can be added though if you think it would be a better addition. I'll probably make a PR if we decide to add it in the future.

As discussed, we'll leave it to be revisited later when we get to this flow. 👍

PR #193, opened 2020-12-02
Add Integration tests using Cucumber tests for Device Controller

mohhef:
# General info

This PR introduces integration tests using cucumber
Cucumber provides a verbose way to understanding integration tests to non-developers
It provides the reusability of some code

**Why use cucumber?**
- gherkin language that provides an easy way for po, pms to understand the integration tests.
- A common way to do integration tests If a test has multiple behaviours that could be added as **examples** in the **.feature** document
- Good when presenting infront of prof(extra cookies)

**Why not use?**
- will our controller benefit from the **examples** section that facilitates have a lot of tests in less developing time.
- pm,po,prof might not care about having verbose integration tests

Note:

The integration tests are running on its own H2 database


src/test/resources/application.properties
```diff
@@ -3,4 +3,13 @@ spring.jpa.hibernate.ddl-auto=update
...
+#CI testing
+spring.datasource.url=jdbc:h2:mem:testdb
+spring.datasource.driverClassName=org.h2.Driver
+spring.datasource.username=sa
+spring.datasource.password=password
+spring.jpa.database-platform=org.hibernate.dialect.H2Dialect
+#Local testing
+#spring.datasource.url=jdbc:mysql://localhost:3306/switchboardtest?useSSL=false&serverTimezone=UTC
+#spring.datasource.username=root
+#spring.datasource.password=root
```
ButlerJAL: Why are you not locally testing with the embedded database the same way we would do in CI? Also, are the changes in this file necessary? Isn't the `spring.datasource.url` property sufficient when using defaults?

ButlerJAL (commented):
I have not used the Cucumber platform before.

From reviewing your code, Cucumber doesn't look to be too complicated to learn. Chances are I won't be able to dive much into a new platform in the short term. If Cucumber helps with integration testing productivity, I'll make a best effort to bring myself up to speed in the medium term.

Are these tests actually testing the entire response or just the response status code? Does import TestRestTemplate take care of testing the entire response along side Cucumber?

I'm all for working with new tools if they bring extra value for our purposes. I'll approve merging these changes if any contributor, including yourself @mohhef, could respond to this comment with a convincing "Cucumber is the way to go".

src/test/resources/application.properties
mohhef: Commented out code could be removed, ill remove it.

mohhef:
> I have not used the Cucumber platform before.
> 
> From reviewing your code, Cucumber doesn't look to be too complicated to learn. Chances are I won't be able to dive much into a new platform in the short term. If Cucumber helps with integration testing productivity, I'll make a best effort to bring myself up to speed in the medium term.
> 
> Are these tests actually testing the entire response or just the response status code? Does import TestRestTemplate take care of testing the entire response along side Cucumber?
> 
> I'm all for working with new tools if they bring extra value for our purposes. I'll approve merging these changes if any contributor, including yourself @mohhef, could respond to this comment with a convincing "Cucumber is the way to go".

You can think of Cucumber as a way to just represent tests, what you can do with cucumber, you can do it without cucumber tests in a different representation.
TestRestTemplate is simply a way to trigger endpoints without actually calling the controller method explicitly. For example, normally, if you would want to trigger a /devices endpoint, you would trigger it by starting up the server and invoking the /devices endpoint from which it will implicitly call the controller. This is what TestRestTemplate is for, it's for invoking endpoints after the server is running. As a side note, RestTestTemplate has nothing to do with Cucumber, it can be used with or without it.
(testresttemplate, resttemplate, restassured are all for the same purpose, if you are familiar with any, in our case I'm using testresttemplate)
I assert the response body when the endpoint returns a 200 status code, otherwise, I just assert the status(e.g. 406)

src/test/java/features/DeviceStepDefinition.java
```diff
@@ -0,0 +1,86 @@
...
+  @When("I update a device")
+  public void i_update_a_device() {
+    DeviceDto deviceDto = DeviceFixture.getDeviceDto();
+    deviceDto.setDisplayName("New display name");
+    HttpEntity httpEntity = new HttpEntity(deviceDto);
+    responseEntity = testRestTemplate.exchange(uri, HttpMethod.PUT, httpEntity, DeviceDto.class);
+  }
+
+  @Then("I should get a response with status {int}")
+  public void i_should_get_a_response_with_status(Integer code) {
```
FGRCL, line 76: Is this reusable accross different .feature classes?

src/test/java/features/DeviceStepDefinition.java
```diff
@@ -0,0 +1,86 @@
...
+    assertEquals(code, responseEntity.getStatusCodeValue());
+  }
+
+  @Then("I should get a response with body")
+  public void i_should_get_a_response_with_body() {
```
FGRCL, line 85: Would it be possible to have a decoder as an input here? Or should we make definitions for all those cases here? I'm starting to think about testing the isRendezvous mode of the StreamModel.

src/test/resources/features/Device.feature
```diff
@@ -0,0 +1,75 @@
...
+    Given I want to create a device
+    When I create a device
+    Then I should get a response with status <code>
+    And I should get a response with body
+
+    Examples:
+      | code |
+      | 200  |
+
+  Scenario Outline: Create Same Device
```
FGRCL: From this test I understand that the database isn't cleared after each test, is that accurate? In my mind it should be to avoid leaky tests, what do you think?

src/test/resources/features/Device.feature
```diff
@@ -0,0 +1,75 @@
...
+    Then I should get a response with status <code>
+
+    Examples:
+      | code |
+      | 409  |
+
+  Scenario Outline: Get Devices
+    Given I want to get all devices
+    When I get all devices
+    Then I should get a response with status <code>
```
FGRCL, line 26: It could be nice to validate the content of the response here

src/test/resources/features/Device.feature
```diff
@@ -0,0 +1,75 @@
...
+  Scenario Outline: Get a Non-existing Device
+    Given I want to get the device with serialNumber <serialNumber>
+    When I get a device
+    Then I should get a response with status <code>
+
+    Examples:
+      | code | serialNumber |
+      | 404  | "2"          |
+      | 404  | "abc"        |
+
```
FGRCL, line 52: There are two lines of examples here, will this run as two "test cases"?

FGRCL (changes requested):
I basically have just one change request for now, I talked with Mohammed about this on Slack. We should find a way to make the integration tests runnable with `mvn test` to integrate nicely with the CI. HI searched a little bit but could find the way to do it. Here's what I tried:
- change `RunCucumberCT.java` to `RunCucumberTest`. maven seaches for classes that match "*Test.java" by default so this might be required
- delcared explicitly the `project.build.testSourceDirectory` property in pom.xml
- added the io.cucumber.cucumber-testng depency to the pom.xml

Appart from that I basically just have questions about how the cucumber tests work. It looks really cool! good initiative.

src/test/resources/application.properties
mohhef: You are right, fixed

mohhef:
> I basically have just one change request for now, I talked with Mohammed about this on Slack. We should find a way to make the integration tests runnable with `mvn test` to integrate nicely with the CI. HI searched a little bit but could find the way to do it. Here's what I tried:
> 
> * change `RunCucumberCT.java` to `RunCucumberTest`. maven seaches for classes that match "*Test.java" by default so this might be required
> * delcared explicitly the `project.build.testSourceDirectory` property in pom.xml
> * added the io.cucumber.cucumber-testng depency to the pom.xml
> 
> Appart from that I basically just have questions about how the cucumber tests work. It looks really cool! good initiative.

Took me good 4 hours to figure the solution. The problem is that cucumber was using different JUnit version than other tests

src/test/java/features/DeviceStepDefinition.java
mohhef: Yes, all step definitions are reusable, however, its better if we reuse corresponding step definitions to its corresponding feature class and not other feature classes. To not introduce confusion

src/test/java/features/DeviceStepDefinition.java
mohhef: Yes you can enter a body as an input. But what I do is I assert the body that is returned by a get request for example with the fixtures that we have. That way I dont need to pass a body explicitly

src/test/resources/features/Device.feature
mohhef: The database is only created once. 
If you create a database after each test:
1. It will be very heavy and it will take long
2. we want to try to simulate a realistic environment where a db is only created once.

src/test/resources/features/Device.feature
mohhef: Yes!! and thats the beauty of cucumber, although my examples don't show this. But basically, you can have a test that runs a lot of examples to test different behaviors:
an example can be found here: https://www.baeldung.com/cucumber-data-tables

src/test/resources/features/Device.feature
mohhef: fixed

FGRCL: I got confused because the code moved up but the comment didn't, oops. Nice work!

FGRCL (dismissed):
I can see that the tests are being run on this pull request already so that's very cool. No more comments on my end!

src/test/java/features/RunCucumberTest.java
```diff
@@ -0,0 +1,6 @@
+package features;
+
+import io.cucumber.junit.platform.engine.Cucumber;
+
+@Cucumber
+public class RunCucumberTest {}
```
felixlapierre: We should rename this class to `CucumberTest` so we avoid having class names that are actions.

felixlapierre (dismissed):
Is the database being pre-populated with entries so that the tests pass, or do the tests rely on the CreateDevice test happening before the Get Devices test so that the items exist in the db? Are the tests independent?

Beyond that, I'm down to give Cucumber a try. I find that it disconnects the test definition a lot from its implementation, but it looks cool and might give us a lot of value when we show this to the professor.

(Tests pass on my end)

src/test/java/features/RunCucumberTest.java
mohhef: No the database is not prepopulated with entries, Its made such  that it inserts in the database before it gets, etc....

felixlapierre (approved):

ButlerJAL (approved):
LGTM

PR #205, opened 2020-12-08
[#200] Registration bug

### LINKED ISSUE #200: Sample apps do not register channels
Steps to reproduce:
- Start the backend/frontend
- Start the sample sender
- Enter device details and click "register"
- Open the frontend
- Click on your sender in the device list page to view its channels

Expected result
- The device has at least one channel

Actual result
- The device has no channels 

Cause:
- The register encoder/decoder endpoint used to take the channel information in the request body and create the appropriate channels. Ever since we added the channel controller, this flow is no longer supported and the channels need to be created in separate requests.

Potential fixes:
- Fix the regression in the registration endpoint. It would be simpler if the encoder/decoder can just send one request that has all of its information. The channel controller will still be useful for getting / deleting / updating specific channels.

mohhef:
# General info

The proper fix is to either do:

1. Only use channel controller to associate encoder/decoder with channels
2. Only use encoder/decoder to persist channels

in that case #1 would be more appropriate since we will be able to associate channels with already existing encoder/decoders with a smaller payload

The fix introduced in this pr, is somewhat a workaround. That allows you do both(persist channels from creatiing an encoder/decoder or persisting channels from the channels entity).

Merge https://github.com/bean-pod/switchboard-samples/pull/8 before merging this pr
**Issue number**: 200

### DIFF
--- a/src/main/java/org/beanpod/switchboard/entity/OutputChannelEntity.java
+++ b/src/main/java/org/beanpod/switchboard/entity/OutputChannelEntity.java
@@ -30,8 +30,7 @@ public class OutputChannelEntity {
   @Column(name = "id")
   private Long id;
 
-  @OneToOne(
-      cascade = {CascadeType.MERGE})
+  @OneToOne(cascade = {CascadeType.MERGE})
   @JoinColumn(name = "channel_id")
   private ChannelEntity channel;
 
--- a/src/main/java/org/beanpod/switchboard/entity/OutputChannelEntity.java
+++ b/src/main/java/org/beanpod/switchboard/entity/OutputChannelEntity.java
@@ -31,14 +31,13 @@ public class OutputChannelEntity {
   private Long id;
 
   @OneToOne(
-      optional = false,
       cascade = {CascadeType.MERGE})
   @JoinColumn(name = "channel_id")
   private ChannelEntity channel;
 
   @ManyToOne(
       fetch = FetchType.LAZY,
-      cascade = {CascadeType.MERGE, CascadeType.PERSIST})
+      cascade = {CascadeType.MERGE})
   @JoinColumn(name = "encoder_serial")
   @JsonIgnoreProperties("output")
   private EncoderEntity encoder;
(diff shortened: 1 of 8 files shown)


src/main/java/org/beanpod/switchboard/entity/OutputChannelEntity.java
```diff
@@ -31,17 +30,14 @@
   @Column(name = "id")
   private Long id;
 
-  @OneToOne(
-      optional = false,
```
FGRCL, line 35: Why are we removing this? We shouldn't have an input or output channel that doesn't have an associated channel object right?

src/main/java/org/beanpod/switchboard/entity/OutputChannelEntity.java
```diff
@@ -31,17 +30,14 @@
...
-      cascade = {CascadeType.MERGE})
+  @OneToOne(cascade = {CascadeType.MERGE})
   @JoinColumn(name = "channel_id")
   private ChannelEntity channel;
 
   @ManyToOne(
-      optional = false,
```
FGRCL, line 41: Same comment here, why are we removing this?

FGRCL (approved):
As far as workarounds go, it's not half bad. Maybe we're due for a second backend bonus meeting to fix this circular reference business at once. What do you think @felixlapierre. For the sonarque failure, I'd agree with you in that it's a "false positive", but I'm not sure what we could do about that. Maybe @MrJCipherButtles has any idea?

src/main/java/org/beanpod/switchboard/entity/OutputChannelEntity.java
mohhef: I agree with you, however if i have the optional tag, I would get an error like: not-null property references a null or transient value.
This is part of the workaround solution where the logic doesn't really meet the code.

src/main/java/org/beanpod/switchboard/entity/OutputChannelEntity.java
mohhef: I agree with you, however if i have the optional tag, I would get an error like: not-null property references a null or transient value.
This is part of the workaround solution where the logic doesn't really meet the code.

felixlapierre (approved):
I've checked out the code locally and everything works. I was able to register a sender and receiver using the sample apps just fine, and creating a stream between them still works as expected.

I agree that this is a symptom of an underlying issue and that we'll need to put some effort in to find a better fix to the circular dependency. I also agree that Sonarcloud is marking the code as duplicate by mistake.

PR #206, opened 2020-12-19
[#161] Update the API contract

### LINKED ISSUE #161: Update existing controllers to use contract / implement api spec
During a meeting we decided to go contract-first for developing the backend, instead of generating the contract from the code.
We have to change existing controllers that aren't using this method so that they implemented the generated API spec

FGRCL:
# General info
This PR is part of the refactor to enforce the API contract. I am updating the api-spec only for now to make pull requests for each controller to it's more manageable

**Issue number**: #161 

**Task description**: 
 - Update the api-spec contract to match the requirements.
 
 

# Testing
No tests required

**Test coverage**:

# Additional notes

**Note to reviewers**:

**To do before merging**:

### DIFF
--- a/contracts/api-spec.yaml
+++ b/contracts/api-spec.yaml
@@ -106,10 +106,11 @@ paths:
                 $ref: '#/components/schemas/EncodersModel'
     post:
       operationId: createEncoder
-      description: Create a new encoder client
+      description: Create a new encoder device
       tags:
         - Encoders
       requestBody:
+        description: The serial number of the device which should be registered as an encoder
         required: true
         content:
           application/json:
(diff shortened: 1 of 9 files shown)


contracts/api-spec.yaml
```diff
@@ -409,8 +458,8 @@ components:
           type: string
           example: 'v7BM3ejS'
 
-    OutputChannelModel:
-      description: A channel through which an encoder outputs a video stream
+    ChannelModel:
+      description: Parent for input and output channels
```
felixlapierre, line 462: I'm kinda iffy on that description but that might just be a consequence of the design being a bit strange

felixlapierre (dismissed):
Looks good so far but you'll have to fix the compilation issues in StreamController which already implements StreamApi

felixlapierre (approved):

ButlerJAL (approved):
Looks good!

PR #207, opened 2020-12-23
[#196] Create logs

### LINKED ISSUE #196: Create / View Logs endpoints
Create a new table in the backend to hold the logs

Note: These logs aren't developer-level logs, these are user-facing logs that are meant to allow users of the app to see a history of what's been happening in the application. Events we plan on logging include:

- A device has registered itself in the system
- Device came online
- Device went offline
- Device started streaming (on channel X to channel Y of device Z)
- Device encountered an error during stream with ID N
- Etc

However, we want to leave the logging system flexible and leave it up to the user to interpret the logs. So we aren't going to store anything about the meaning of a log, just the string of the log message.

Logs should have the following properties
- The time at which the log was generated. The server should handle determining the time, it shouldn't be in the body of the create log request
- The device that created the log
- The log "level". Leave this as a string, but the possible values will probably only include INFO and ERROR. Since these aren't dev-facing logs there will be no DEBUG log level.
- The log message (also a string). We might need to do input validation on this to protect against injection.
- We will probably need to generate a unique ID for each log
If you think there could be other information in here let me know.

** Acceptance Criteria **
- POST endpoint that creates a log, that takes the device serial, the log level, and the log message. This endpoint should sanitize the input to protect against database injection.
- GET endpoint that returns all logs in the system
- GET endpoint that takes the serial number of a device and returns all the logs of that device

mohhef:
# General info

- Put logs into db when a device is online or offline
- put logs when starting a stream

**Note to reviewers**:
- This pr uses AOP to achieve its functionality

**To do before merging**:
- [x] unit tests
i did not unit test the aspect class, I do not know if it logically makes sense to unit test it. I think its something to be handled with integration tests


src/main/java/org/beanpod/switchboard/repository/InputChannelRepository.java
```diff
@@ -2,7 +2,9 @@
 
 import org.beanpod.switchboard.entity.InputChannelEntity;
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
 
+@Repository
 public interface InputChannelRepository extends JpaRepository<InputChannelEntity, Long> {
```
felixlapierre, line 8: why did we need to introduce this change? was the code not working before?

ButlerJAL (dismissed):
I had a quick one-on-one with @mohhef to go over his PR. It's not everyday I think of AOP. It's a very fitting solution here. And, the code looks great!

contracts/api-spec.yaml
```diff
@@ -365,6 +376,29 @@ paths:
...
+          example: '123'
+        dateTime:
+          type: string
+          format: date-time
+        message:
+          type: string
+          example: 'A log message'
+        level:
+          type: string
+          example: 'info'
```
felixlapierre, line 400: We also need to know what device created the log. This will be useful so we can view device-specific logs. That requirement isn't captured in this implementation IMO but i'm willing to merge this and add it later.

src/main/java/org/beanpod/switchboard/aop/DeviceAspect.java
```diff
@@ -0,0 +1,49 @@
...
+import org.openapitools.model.DeviceModel;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Aspect
+@Configuration
+@Component
+@RequiredArgsConstructor
+public class DeviceAspect {
```
felixlapierre, line 19: This is a perfect place to use aspects, good job!! 👍 🔥

src/main/java/org/beanpod/switchboard/aop/DeviceAspect.java
```diff
@@ -0,0 +1,49 @@
...
+    Object[] args = joinPoint.getArgs();
+    CreateDeviceRequest createDeviceRequest = (CreateDeviceRequest) args[0];
+    String status = createDeviceRequest.getStatus();
+    String serialNumber = createDeviceRequest.getSerialNumber();
+    String message = String.format("Device with serial number %s has been created", serialNumber);
+    if (status.equals("online") || status.equals("offline")) {
+      message = message.concat(" with status " + status);
+    }
+    logService.createLog(message, "info");
+  }
```
felixlapierre: i'd recommend adding some whitespace to break up these lines into distinct blocks, it's a bit hard to read here

maybe add an empty line after lines 27, 33, and maybe another? whatever you prefer. same for the next method.

src/main/java/org/beanpod/switchboard/aop/DeviceAspect.java
```diff
@@ -0,0 +1,49 @@
...
+      "execution(* org.beanpod.switchboard.controller.DeviceController.updateDevice(..))")
+  public void updateDevice(JoinPoint joinPoint) {
+    String message;
+    Object[] args = joinPoint.getArgs();
+    DeviceModel deviceModel = (DeviceModel) args[0];
+    String status = deviceModel.getStatus();
+    if (status.equals("online") || status.equals("offline")) {
+      message = "A device has been updated with a status of " + status;
+      logService.createLog(message, "info");
+    }
```
felixlapierre: We actually aren't using the status that's in the database, we determine the status in the frontend based on the time of last communication. That being said, doing it like that makes it impossible to determine when a device goes offline or online.... i think we leave this for now and i'll have to figure something else out later.

felixlapierre (dismissed):
I have 1 question and 1 whitespace comment, the others are just me thinking aloud about the way i designed the requirements for this feature. Also haven't tested it locally but lgtm!

src/main/java/org/beanpod/switchboard/repository/InputChannelRepository.java
mohhef: No it was working fine, it is just there to make it clear that it is a repo component

contracts/api-spec.yaml
mohhef: That i put it in the log message. I had a reason to do it that way but i kind of forgot why, ill reiterate over my thinking next sprint and see. It makes sense to have it in its separate column

mohhef: e.g. Device with serial number 4 has been created. 4 is orginally a var

src/main/java/org/beanpod/switchboard/aop/DeviceAspect.java
mohhef: Sounds good we can discuss this next sprint

src/main/java/org/beanpod/switchboard/aop/DeviceAspect.java
mohhef: fixed

contracts/api-spec.yaml
felixlapierre: yeah, but we'll need to represent it in the database entry too so we can query for a device's logs. agreed, we'll take a look again next sprint

felixlapierre (approved):
My comments were addressed, LGTM

ButlerJAL (approved):

PR #211, opened 2021-01-09
[#192] Delete stream in frontend

gdzooks:
# General info

**Issue number**: 192

**Task description**: 
- For each stream in the view streams table, a delete button (such as a garbage can icon) is visible
- When the delete button is clicked, a popup is shown that reads "Are you sure you want to end the stream?"
- If the user clicks cancel, the popup closes and nothing happens
- If the user clicks yes, the delete stream endpoint is called, and the page in the frontend is updated to show that the stream doesn't exist anymore
 
 
| Original   | Updated|
|:---------------:|:---------------:|
| image | image |

# Testing
**Test coverage**:

* 100%! 
* Entire test suite for `StreamList` needs to be rewritten to use Enzyme but I don't think this is the issue/PR to do it in.

# Additional notes

**Note to reviewers**:

Couldn't get delete working with `SampleData` written the way it is - the whole class needs a big refactor if we want creating, modifying, and deleting to work with it (creating streams doesn't have a sample function either).
I don't have backend and/or samples setup so I didn't actually test that `axios.delete` worked as expected so I'm counting on one of you~

**To do before merging**:
N/A

Issue #192, opened 2020-12-01
Delete a stream in the frontend


felixlapierre:
**Acceptance Criteria**
- For each stream in the view streams table, a delete button (such as a garbage can icon) is visible
- When I click on the delete button, a popup is shown "Are you sure you want to end the stream?"
- If the user clicks cancel / no, the popup closes and nothing happens
- If the user clicks yes, the delete stream endpoint is called, and the page in the frontend is updated to show that the stream doesn't exist anymore

PR #215, opened 2021-01-10
Allow CORS on delete endpoints

felixlapierre:
# General info

**Issue number**: Issue discovered as part of issue #192 while testing PR #211

**Task description**: 
- Enable CORS for DELETE requests 

# Testing
N/A, this change is in configuration that cannot be covered by unit tests & will eventually be covered by the system test for deleting a stream.

### DIFF
--- a/src/main/java/org/beanpod/switchboard/config/WebMvcConfiguration.java
+++ b/src/main/java/org/beanpod/switchboard/config/WebMvcConfiguration.java
@@ -12,10 +12,7 @@ public class WebMvcConfiguration {
     return new WebMvcConfigurer() {
       @Override
       public void addCorsMappings(CorsRegistry registry) {
-        registry
-            .addMapping("/**")
-            .allowedOrigins("*")
-            .allowedMethods("GET", "POST", "PUT", "DELETE");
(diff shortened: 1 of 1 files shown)


src/main/java/org/beanpod/switchboard/config/WebMvcConfiguration.java
```diff
@@ -12,7 +12,10 @@ public WebMvcConfigurer webMvcConfigurer() {
     return new WebMvcConfigurer() {
       @Override
       public void addCorsMappings(CorsRegistry registry) {
-        registry.addMapping("/**");
+        registry
+            .addMapping("/**")
+            .allowedOrigins("*")
```
ButlerJAL: This line is causing a security hotspot. It's too permissive. Do you want to fix by explicitly defining allowed origins? Or, won't fix?

felixlapierre: Fix still works without the `allowedOrigins` line so i removed it, the security hotspot should be gone

ButlerJAL (approved):

mohhef (approved):
Looks good

PR #244, opened 2021-01-17
[#241] Update device model

### LINKED ISSUE #241: Update device model in frontend to match differentiation in backend
In the backend devices are separated into encoder and decoder (sender and receiver), but the frontend model has no property to reflect this. We need to be able to easily differentiate between device types at several points in the frontend and passing an extra prop based on the device tables is... a bad idea. Just in general.

~~I will not be taking criticisms on this refactor, thank you.~~

gdzooks:
# General info

**Issue number**: #241 

**Task description**: 

 - Update the device model in the frontend to have a `deviceType` field that will be used to differentiate between senders & receivers at the model level, like in the backend.

 
| Original   | Updated|
| N/A | N/A |

# Testing

**Test coverage**:
- unchanged

# Additional notes

**Note to reviewers**:
- N/A

**To do before merging**:
- N/A

### DIFF
--- a/frontend/src/api/DeviceApi.js
+++ b/frontend/src/api/DeviceApi.js
@@ -47,8 +47,8 @@ export function getSenders(callback) {
             sender.device.displayName,
             getStatus(sender.lastCommunication),
             channels,
-            ["Additional Device details go here"],
-            "encoder"
+            "encoder",
+            ["Additional Device details go here"]
           );
         })
       );
@@ -87,8 +87,8 @@ export function getReceivers(callback) {
             receiver.device.displayName,
             getStatus(receiver.lastCommunication),
             channels,
-            ["Additional Device details go here"],
-            "decoder"
+            "decoder",
+            ["Additional Device details go here"]
           );
         })
       );
--- a/frontend/src/api/DeviceApi.js
+++ b/frontend/src/api/DeviceApi.js
@@ -47,7 +47,8 @@ export function getSenders(callback) {
             sender.device.displayName,
             getStatus(sender.lastCommunication),
             channels,
-            ["Additional Device details go here"]
+            ["Additional Device details go here"],
+            "encoder"
           );
         })
       );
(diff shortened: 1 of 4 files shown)


sharon-ho (dismissed):
LGTM! 👍

frontend/src/api/DeviceApi.js
```diff
@@ -47,7 +47,8 @@ export function getSenders(callback) {
             sender.device.displayName,
             getStatus(sender.lastCommunication),
             channels,
-            ["Additional Device details go here"]
+            ["Additional Device details go here"],
+            "encoder"
```
jun-loke: Any particular reason for this order for the arguments?
Personally, I think that "extras" should be the last argument in the list, since it isn't actually being used yet and isn't a "real" value, whereas the "encoder" value is actually useful.

jun-loke (changes requested):
Still need to run it on my computer (Nothing to do with this code, computer troubles. Had to wipe the whole thing) but there's small question that I had. Once we discuss it I'll approve

frontend/src/api/DeviceApi.js
gdzooks: none in particular, i just wasn't super sure where else to put it... immediately before extras?

jun-loke: yeah, I think it would be better to have it before extras. Since, if we remove extras, which has been mentioned as an option, it'll be easier if it's at the end

felixlapierre: +1 Agree with Jun, after that change is made I can give my approval aswell

gdzooks: moved it up one! b2e1224

frontend/src/api/SampleData.js
```diff
@@ -120,7 +129,8 @@ export function getSampleSender() {
     "Sender 1",
     "Online",
     sampleOutputChannels,
-    extras
+    extras,
+    "encodr",
```
jun-loke: Typo & order and comma

jun-loke (changes requested):
Smol typo

frontend/src/api/SampleData.js
gdzooks: OUF thanks for catching that! fixed 9ea91ea

jun-loke (approved):
Delicious. Very cash money

sharon-ho (approved):
Looking good!

PR #256, opened 2021-01-21
[#229] Add jmeter load tests to all endpoints

### LINKED ISSUE #229: Performance Testing
Record service logs using logging libraries, and perform Load and Stress testing. Analyze the collected data to find ways to improve the server performance

- Add logs in the service that track performance
- Set up Jmeter locally to test against deployed service on Heroku
- Record results for a load of 100 users with 10 devices each
- Analyze the results and highlight improvement opportunities

mohhef:
# General info

- Added jmeter load test to all endpoints.
- To efficiently test the jmeter tests, download apache jmeter and import the jmx file, then use the aggregate report/graph to see the average response times
- The goal is to save aggregated results after each sprint and try to improve on the response times between each sprint
- Make sure to set foreign keys to 0 when running delete tests

To easily truncate all tables on heruko:
```
SET FOREIGN_KEY_CHECKS = 0; 
TRUNCATE table encoder; 
TRUNCATE table decoder; 
TRUNCATE table device; 
TRUNCATE table channel; 
TRUNCATE table input_channel; 
TRUNCATE table output_channel; 
SET FOREIGN_KEY_CHECKS = 1;

```
**Issue number**: 229


FGRCL (dismissed):
It ran O.K. on my machine, but I got 71.85% of request errors when running it. Is that expected?

mohhef:
> It ran O.K. on my machine, but I got 71.85% of request errors when running it. Is that expected?

Good catch, I fixed a few typos. The error percentage should be <=50, the reason for that is because i am also testing incorrect requests, so it is most likely never going to be 0.
Make sure to delete the table contents before running the tests, other wise the post requests might give you errors

FGRCL (dismissed):
Yup! I got less errors this time

felixlapierre:
I was able to run it just fine on my end! I think we should get the load test on create stream completed before we merge, so as soon as that's done i'm good to go.

felixlapierre (approved):
The error rate has gone way down and creating streams works! I'm ready to give this a LGTM! And deleting streams will work now that a fix has been merged to master.

FGRCL (approved):

PR #257, opened 2021-01-23
[#229] Service logs

### LINKED ISSUE #229: Performance Testing

mohhef:
# General info

- Adds simple logs in the service.
- Generation method of generating IDs has changed to Identity, that will allow when truncating the tables, the IDs to reset to 0 and not continue from the previously persisted value.

If you encounter any errors, make sure to recreate the database.

**Issue number**: 229

### DIFF
--- a/src/main/java/org/beanpod/switchboard/service/DecoderServiceImpl.java
+++ b/src/main/java/org/beanpod/switchboard/service/DecoderServiceImpl.java
@@ -2,6 +2,7 @@ package org.beanpod.switchboard.service;
 
 import java.util.List;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.beanpod.switchboard.dao.DecoderDaoImpl;
 import org.beanpod.switchboard.dao.StreamDaoImpl;
 import org.beanpod.switchboard.dto.DecoderDto;
@@ -12,6 +13,7 @@ import org.springframework.stereotype.Component;
 
 @Component
 @RequiredArgsConstructor
+@Slf4j
 public class DecoderServiceImpl implements DecoderService {
   private final StreamDaoImpl streamDao;
   private final DecoderDaoImpl decoderDao;
@@ -19,6 +21,7 @@ public class DecoderServiceImpl implements DecoderService {
 
   @Override
   public List<StreamDto> getDecoderStreams(String decoderSerialNumber) {
+    log.info("Getting Decoder's {} streams", decoderSerialNumber);
     DecoderDto decoder =
         decoderDao
             .findDecoder(decoderSerialNumber)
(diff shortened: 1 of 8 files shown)


src/main/java/org/beanpod/switchboard/service/EncoderServiceImpl.java
```diff
@@ -12,20 +13,23 @@
...
   public List<StreamDto> getEncoderStreams(String encoderSerialNumber) {
+    log.info("Getting Encoder's {} streams", encoderSerialNumber);
     EncoderDto encoderDto =
         encoderDao
             .findEncoder(encoderSerialNumber)
             .orElseThrow(() -> new ExceptionType.DeviceNotFoundException(encoderSerialNumber));
 
     encoderDto.setLastCommunication(dateUtil.getCurrentDate());
     encoderDao.save(encoderDto);
+    log.debug("Updated encoder's last communication date");
```
felixlapierre: This log should also include the encoder serial number since, due to asynchronous execution, it might not appear right after the previous one so we won't know which encoder got updated. Same for decoder.

felixlapierre (dismissed):
One small suggestion otherwise LGTM

src/main/java/org/beanpod/switchboard/service/DecoderServiceImpl.java
```diff
@@ -2,6 +2,7 @@
 
 import java.util.List;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
```
gdzooks, line 5: can i ask what this import does? 😅

mohhef: Yes, it imports slf4j which is a logging library in Java

gdzooks: ahh it generates the log instance! i get it now!

gdzooks (dismissed):
felix's comment makes sense to me, so i'm down to re-approve quickly if you make that change! otherwise looks good to me, thanks for answering my question :)

src/main/java/org/beanpod/switchboard/service/EncoderServiceImpl.java
mohhef: fixed

gdzooks (approved):
lgtm! :tada:

felixlapierre (approved):
Re-approve since comments addressed LGTM

PR #264, opened 2021-01-27
[#228] Create system tests for existing controllers

### LINKED ISSUE #228: Create system tests for existing controllers
Create system tests for 
- adding a decoder
- creating a stream
- viewing the stream
- deleting the stream

ButlerJAL:
# General info

Create system tests using Selenium

**Issue number**: #228 

**Task description**: 
 - adding a decoder
 - creating a stream
 - viewing the stream
 - deleting the stream

# Testing

**Test coverage**: n/a

# Additional notes

**Note to reviewers**:
 - [Download](https://chromedriver.chromium.org/downloads) and move chromedriver into `src/test/java/system_tests/` (You may need to give permission to run chromedriver on your OS)
 - chromedriver expects that you have the corresponding chrome browser installed on your system.
 - Run back-end and front-end locally, preferably your database should be empty before running the system test.

**To do before merging**: n/a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ build/
 
 ### Node modules###
 node_modules/
+/src/test/java/system_tests/chromedriver*
--- a/src/test/java/system_tests/TrialSeleniumTest.java
+++ b/src/test/java/system_tests/TrialSeleniumTest.java
@@ -50,13 +50,19 @@ public class TrialSeleniumTest {
 
     // Set up Selenium Chrome Driver
     if (System.getProperty("os.name").toLowerCase().contains("windows")) {
-      System.setProperty("webdriver.chrome.driver", "src\\test\\java\\system_tests\\chromedriver.exe");
+      System.setProperty("webdriver.chrome.driver",
+          "src\\test\\java\\system_tests\\chromedriver.exe");
     } else {
       System.setProperty("webdriver.chrome.driver", "src/test/java/system_tests/chromedriver");
     }
     driver = new ChromeDriver();
     driver.manage().window().setSize(new Dimension(1280, 1024));
     driver.manage().timeouts().implicitlyWait(15, TimeUnit.SECONDS);
+  }
+
+  @AfterAll
+  static void tearDown() {
+    driver.quit();
   }
 
   @Test
@@ -74,7 +80,8 @@ public class TrialSeleniumTest {
     List<WebElement> devicesRows =
         encodersTable.findElements(By.tagName("tr")); // find all tr elements inside found table
     boolean assertValue =
-        devicesRows.stream().anyMatch(row -> row.getText().contains(EncoderFixture.getEncoderEntity1().getDevice().getDisplayName()));
+        devicesRows.stream().anyMatch(row -> row.getText()
+            .contains(EncoderFixture.getEncoderEntity1().getDevice().getDisplayName()));
 
     assertTrue(assertValue);
   }
@@ -95,7 +102,8 @@ public class TrialSeleniumTest {
     List<WebElement> devicesRows =
         decodersTable.findElements(By.tagName("tr")); // find all tr elements inside found table
     boolean assertValue =
-        devicesRows.stream().anyMatch(row -> row.getText().contains(DecoderFixture.getDecoderEntity2().getDevice().getDisplayName()));
+        devicesRows.stream().anyMatch(row -> row.getText()
+            .contains(DecoderFixture.getDecoderEntity2().getDevice().getDisplayName()));
 
     assertTrue(assertValue);
   }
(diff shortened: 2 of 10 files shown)


src/test/java/system_tests/TrialSeleniumTest.java
```diff
@@ -1,73 +1,208 @@
...
+  @Test
+  @Order(3)
+  void testCreateStream() {
+    driver.get("http://localhost:3000/Streaming");
+    {
+      WebDriverWait wait = new WebDriverWait(driver, 100);
+      wait.until(
+          ExpectedConditions.elementToBeClickable(
+              By.cssSelector("#SenderTable .MuiList-root .MuiSvgIcon-root")));
+    }
```
FGRCL: Any reason why we're using code blocks here? If it's just for readability I think that would be a good opportunity to add a helper method with a name.

.gitignore
```diff
@@ -33,4 +33,5 @@ build/
 .vscode/
 
 ### Node modules###
-node_modules/
\ No newline at end of file
+node_modules/
+/src/test/java/system_tests/chromedriver*
```
FGRCL, line 37: Why not commit the chrome driver to version control, is it for chrome version compatibility issues?

src/test/java/system_tests/TrialSeleniumTest.java
```diff
@@ -1,73 +1,208 @@
...
+    driver.findElement(By.cssSelector(".MuiMenuItem-root")).click();
+    driver.findElement(By.cssSelector("#ReceiverTable .MuiList-root .MuiSvgIcon-root")).click();
+    {
+      WebElement element;
+      WebDriverWait wait = new WebDriverWait(driver, 100);
+      element =
+          wait.until(
+              ExpectedConditions.elementToBeClickable(
+                  By.xpath(
+                      "//div[@id='ReceiverTable']/div[3]/ul/div[2]/div/div/div/li/div[2]/div")));
```
FGRCL: I don't think there's much we can do about that, but that is a mouthful

src/test/java/system_tests/TrialSeleniumTest.java
```diff
@@ -1,73 +1,208 @@
...
+      WebElement element;
+      WebDriverWait wait = new WebDriverWait(driver, 100);
+      element =
+          wait.until(ExpectedConditions.elementToBeClickable(By.cssSelector(".MuiMenuItem-root")));
+      Actions builder = new Actions(driver);
+      builder.moveToElement(element).release().perform();
+    }
+    driver.findElement(By.cssSelector("body")).click();
+    driver.findElement(By.cssSelector(".MuiMenuItem-root")).click();
+    driver.findElement(By.cssSelector(".buttonText")).click();
```
FGRCL: I see that this test doesn't have any asserts, could we call the `/stream` endpoint and check that it doesn't return an empty list maybe?

FGRCL (commented):
The test ran well on my end! I have a couple questions on the code though.

.gitignore
ButlerJAL: It’s OS dependent.

src/test/java/system_tests/TrialSeleniumTest.java
ButlerJAL: Nope not much to do about it

src/test/java/system_tests/TrialSeleniumTest.java
ButlerJAL: This is done to have the variable local to the code block it’s defined within

src/test/java/system_tests/TrialSeleniumTest.java
ButlerJAL: Yeah, it’s because the create stream and view stream are two separate tests, but I think they might as well be combined into one. The view test is pretty much the assert of the create test.

.gitignore
FGRCL: We should be able to use different file names though, no? I saw that we already have code that checks to OS an loads the right chromedriver.

src/test/java/system_tests/TrialSeleniumTest.java
felixlapierre: I wouldn't mind combining them into one test

felixlapierre (dismissed):
LGTM. T'was a beauty to watch the system tests automatically clicking through our app.

A few minor suggestions for improvements:
- Is there a way to split our tests into multiple files? (Perhaps by page, or by user story). Or is it not worth it to have to rewrite the setup method each time?
- If we keep everything in one file we should probably rename it from `TrialSeleniumTest` to `SystemTests` or something.
- Add instructions to the readme for "download chromedriver.exe and place it in `src/test/java/system_tests`"

.gitignore
ButlerJAL: Also you got to ensure that the chromedriver matches the chrome version installed on your system. I don’t think chromedriver is something you would want to version control as it’s user specific.

ButlerJAL:
@felixlapierre I think the system test is best kept in one file. I wasn’t a fan of the system test file name either. I’ll rename the file and update the readme as per your suggestion as soon as I can.

FGRCL (dismissed):

felixlapierre (approved):
Comments addressed, LGTM!

FGRCL (approved):

PR #253, opened 2021-01-19
Styling and layout changes to DeviceDetailsPage and minor changes to DeviceList

sharon-ho:
# General info

**Issue number**: N/A

**Task description**: 
- Removing redundant Overview table/tab from the rightmost table in the page
- Add an overview tab to the leftmost table and remove the click functionality to the tab
- Change edit button into an Edit icon and place it next to the Device Name, since it will be the only editable property
- Move Delete button to be on the same line as Device Name
- Extremely minor changes: changed indicator color in DeviceList vertical tabs and added padding to the DetailsPanel in DeviceList
 
 
| Original   | Updated|
|:---------------:|:---------------:|
| image | image |

# Testing

**Test coverage**: 100%

# Additional notes

**Note to reviewers**:
N/A

**To do before merging**:
- [x] Raise code coverage
- [x] Wait for #248 to be merged first

PR #276, opened 2021-01-28
Add delete device functionality to actions menu

gdzooks:
# General info

**Issue number**: 

**Task description**: 
 - As in the title, added delete device functionality to the `Delete` option in the devicelist's `ActionsMenu`
 
 
| Original   | Updated|
| image | image |

# Testing
**Test coverage**:
- 100% lol it's basically the same tests as before xD

# Additional notes

**Note to reviewers**:
- N/A

**To do before merging**:
- [x] wait for #253 to be merged to resolve inevitable import conflicts xD


frontend/src/deviceDetailsPage/DeviceDetailsPage.jsx
```diff
@@ -41,7 +41,7 @@ export default function DeviceDetailsPage(props) {
             </Button>
           </Grid>
           <Grid item xs={2}>
-            <DeleteDeviceButton deleteId={device.serialNumber} />
+            <DeleteDeviceButton button deleteId={device.serialNumber} />
```
sharon-ho: I'm assuming that the button's boolean value is true by default?

frontend/src/general/DeleteDeviceButton.jsx
```diff
@@ -23,14 +40,15 @@ export default function DeleteDeviceButton(props) {
...
   };
 
   return (
     <>
-      <Button id="deleteBtn" variant="outlined" onClick={openDeleteDialog}>
-        Delete
-      </Button>
+      {button
+        ? renderDeleteButton(openDeleteDialog)
+        : renderDeleteMenuItem(openDeleteDialog)}
```
sharon-ho, line 60: I really like how you did this to avoid code duplication by applying conditional rendering! ❤

frontend/src/general/__tests__/DeleteDeviceButton.test.jsx
```diff
@@ -33,7 +35,7 @@ describe("DeleteButton", () => {
   beforeEach(() => {
     handleClick.mockImplementation((open) => [open, setOpen]);
     wrapper = Enzyme.shallow(
-      <DeleteDeviceButton deviceType="encoder" deleteId="1:10:111:999" />
+      <DeleteDeviceButton button deleteId="1:10:111:999" />
```
sharon-ho, line 46: I feel like there should also be a test where button is set as false and the test should check that it renders a MenuItem instead of a Button.

sharon-ho (changes requested):
Just a small question and a minor change requested, otherwise good work!

frontend/src/deviceDetailsPage/DeviceDetailsPage.jsx
gdzooks: yep! and if i try to write `button={true}` the linter tells me to go to hell :)

frontend/src/general/__tests__/DeleteDeviceButton.test.jsx
gdzooks: i thought that test would end up being more trouble than it's worth considering the code coverage is already at 100% but it only took like a minute LOL

frontend/src/deviceDetailsPage/DeviceDetailsPage.jsx
felixlapierre: "yep, this do be a button!!!!"

felixlapierre (approved):
I was having issues with this PR when using it with the backend, but I created a hotfix which @gdzooks has now merged in, so this is a LGTM on my end.

I suggest you fix the code smells in the tests if possible!

gdzooks:
> I suggest you fix the code smells in the tests if possible!

i've tried to write the tests with `.toBe(true)` or `.toBe(false)` and tried changing `setOpen` to `open` to avoid using `.toBeTruthy` and such, but both break the tests. i'm pretty sure it's because of how react mocks states, so unfortunately i don't think i can get rid of these smells. not anytime soon, at least :(

sharon-ho (approved):
Looking good!

PR #293, opened 2021-01-30
Fix for delete device in actions

felixlapierre:
@gdzooks This is a PR into your branch so you can review my fix

* Fixed URL of delete axios request to use environment variables
* Delayed redirect/refresh until a response is received from the server in order to ensure the delete call gets sent & that the device is deleted before we refresh.
* Added a conditional making it so if we are on the device details page, we redirect to the device list page, but if we are on the device list page, we refresh the page. This avoided an issue where it was refreshing twice with my fix when deleting on device list page (one redirect + one refresh)

### DIFF
--- a/frontend/src/api/DeviceApi.js
+++ b/frontend/src/api/DeviceApi.js
@@ -98,9 +98,8 @@ export function getReceivers(callback) {
     });
 }
 
-export function deleteDevice(deviceId, callback) {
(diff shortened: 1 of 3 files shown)


gdzooks (approved):
EYY lgtm! tysm for the fix :sparkles::tada:

PR #294, opened 2021-01-30
Fixes for redirect on stream delete

### LINKED ISSUE #292: Delete stream returns 500 error despite stream being successfully deleted
Steps to reproduce:
- Register a sender and a receiver in the system
- Create a stream between the two devices
- Send a DELETE request to /stream/{id}

Expected result:
- The stream is deleted and a 200 response is received

Actual result:
- The stream is deleted and a 500 response is received

However if you do a GET /stream you will see that the stream was actually deleted.

Cause:
- In StreamAspect after deleting stream, in order to create the log, we get the stream by ID so we can see what encoder/decoder was involved. However, this causes a crash, since the stream was already deleted so it is not found in the database. The resulting exception causes a 500 error.

felixlapierre:
@gdzooks 

* Fixed an issue where the backend returns 500 on deleting stream.
* Redirect now occurs only after the response is received from the backend. 
* ~Added a refresh when creating stream (bonus!)~ Removed this since it was too time-consuming to make the tests pass

Closes #292

### DIFF
--- a/frontend/src/createStream/StreamingTable.jsx
+++ b/frontend/src/createStream/StreamingTable.jsx
@@ -1,5 +1,6 @@
 import React from "react";
 import { Grid } from "@material-ui/core";
+import { withRouter } from "react-router-dom";
 
 import axios from "axios";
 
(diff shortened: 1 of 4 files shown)


felixlapierre:
too tired to fix failing tests today, will fix tmrw

gdzooks (approved):
wondering if you still get a double refresh currently with both `push` and `go` but that's something we can fix in the future if necessary!

felixlapierre:
No double refresh since it's already on the stream page - pushing the page you are on doesn't do anything.
The double refresh happened when i was pushing a different page, then doing `go(0)`, it would go to the other page then immediately refresh

PR #295, opened 2021-01-30
[#217] Get logs of a specific device

### LINKED ISSUE #217: Endpoint to get all logs associated with a device
When viewing a device's details, we want to be able to see all the logs related to a device. For this we need a GET endpoint that takes the serial number of a device and returns a list of logs related to that device.

Right now, the id of the device that generated the log is in the message string, which is not easy to query. We want to easily be able to query which device made the log. Note that we also need a log to be generated by two devices (such as when logging the creation of a stream).

sebhani:
**Issue number**: #217 

**Task description**: 
 - Added an endpoint to retrieve the logs of a specific device
 - Added a unit test for the controller

# Additional notes
Assuming that the device serial number that we would like to find the logs for is X, serial_number attribute in LogEntity can be populated as one the following three cases: 1- X  2- X,Y,  3-Y,X.

The first case appears after creating and updating a device.
The last two cases appear after creating a stream since we have encoder and decoder (not only a single device).


frontend/src/deviceDetailsPage/DeviceDetailsConciseRow.jsx~97be9f2ed042e6a9828aaf93169bb0ad69ee0cc3
```diff
@@ -0,0 +1,72 @@
+import React from "react";
+import Proptypes from "prop-types";
+import {TableContainer, TableCell, TableRow} from "@material-ui/core";
+import ChannelDetailsTable from "../devicelist/ChannelDetailsTable";
+import InputChannelInfo from "../model/InputChannelInfo";
```
jun-loke, line 5: Weird "duplicate" file with random string appended to file name

jun-loke (changes requested):
Seems like you ran the backend linter on the frontend folder, causing a bunch of changes also did something that generated a bunch of weird duplicate files. 
Please revert changes to frontend folder

PR #296, opened 2021-01-30
[#217] Get logs of a specific device - updated PR

### LINKED ISSUE #217: Endpoint to get all logs associated with a device

sebhani:
**Issue number**: Closes #217 

**Task description**: 

- Added an endpoint to retrieve the logs of a specific device
- Added a unit test for the controller

# Additional notes

Assuming that the device serial number that we would like to find the logs for is X, serial_number attribute in LogEntity can be populated as one the following three cases: 1- X 2- X,Y, 3-Y,X.

The first case appears after creating and updating a device.
The last two cases appear after creating a stream since we have encoder and decoder (not only a single device).

### DIFF
--- a/src/main/java/org/beanpod/switchboard/dao/LogDaoImpl.java
+++ b/src/main/java/org/beanpod/switchboard/dao/LogDaoImpl.java
@@ -17,4 +17,8 @@ public class LogDaoImpl {
   public List<LogModel> getLogs() {
     return logMapper.toLogModels(logRepository.findAll());
   }
+
+  public List<LogModel> getDeviceLogs(String serialNumber) {
+    return logMapper.toLogModels(logRepository.findBySerialNumber(serialNumber));
+  }
 }
--- a/src/main/java/org/beanpod/switchboard/repository/LogRepository.java
+++ b/src/main/java/org/beanpod/switchboard/repository/LogRepository.java
@@ -1,11 +1,10 @@
 package org.beanpod.switchboard.repository;
 
+import java.util.List;
 import org.beanpod.switchboard.entity.LogEntity;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.stereotype.Repository;
-
-import java.util.List;
 
 @Repository
 public interface LogRepository extends JpaRepository<LogEntity, Long> {
--- a/src/main/java/org/beanpod/switchboard/repository/LogRepository.java
+++ b/src/main/java/org/beanpod/switchboard/repository/LogRepository.java
@@ -1,9 +1,11 @@
 package org.beanpod.switchboard.repository;
 
-import java.util.List;
 import org.beanpod.switchboard.entity.LogEntity;
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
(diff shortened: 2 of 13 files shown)


felixlapierre (dismissed):
Tested it out on my end and everything is good!

I like the chosen solution for the cases where logs belong to 1 or 2 devices. Maybe we can find a more computationally efficient solution but I like this for the time being.

src/main/java/org/beanpod/switchboard/repository/LogRepository.java
```diff
@@ -11,4 +12,10 @@
   LogEntity save(LogEntity logEntity);
 
   List<LogEntity> findAll();
+
+  // The Query covers three cases: 1- serialNumber = X,Y 2- serialNumber = Y,X 3- serialNumber = X
+  @Query(
+      "SELECT l FROM LogEntity l WHERE l.serialNumber LIKE CONCAT(?1,',%') "
+          + "OR l.serialNumber LIKE CONCAT('%,',?1) OR l.serialNumber = ?1")
+  List<LogEntity> findBySerialNumber(String serialNumber);
 }
```
mohhef, line 21: i think you can use the already defined like or contains method,
https://www.baeldung.com/spring-jpa-like-queries

mohhef (changes requested):
I had just one comment, othewise looks good to me.

src/main/java/org/beanpod/switchboard/dao/LogDaoImpl.java
```diff
@@ -17,4 +17,8 @@
   public List<LogModel> getLogs() {
     return logMapper.toLogModels(logRepository.findAll());
   }
+
+  public List<LogModel> getDeviceLogs(String serialNumber) {
+    return logMapper.toLogModels(logRepository.findBySerialNumber(serialNumber));
+  }
```
mohhef, line 23: i think you forgot to test this

mohhef (changes requested):

src/main/java/org/beanpod/switchboard/repository/LogRepository.java
sebhani: We got three cases for the serial number format, so I don't think we can match it using a single wildcard. Also, I haven't gone with the contain method since it could produce bugs. Let's say you want to match logs with serial number 1, then 13,4 is matched as well since it contains 1.

src/main/java/org/beanpod/switchboard/dao/LogDaoImpl.java
sebhani: Good catch! I just added the test

felixlapierre (approved):
Re-approve

mohhef (approved):
Looks good to me!

src/main/java/org/beanpod/switchboard/repository/LogRepository.java
mohhef: that makes sense

PR #304, opened 2021-02-04
[#218] Create a log when a device status is updated

### LINKED ISSUE #218: Create log when device goes online or offline
When a device goes online or offline, we should create a log to indicate it. The log will have the timestamp, message, log level must be info, device associated with it.

We will need to update the backend code to actually keep the device status field up to date. Right now, we determine device status in the frontend based on the value of the `lastCommunication` field at the time that the device info is retrieved. We would have to generate this log only once when the last communication is more than 10 minutes ago.

This is one solution I propose, but it might not be the best one
- Periodically check all devices in the database. If status is online and lastCommunication is more than 10m ago, set status to offline and generate a log. Instead of checking periodically we could instead check at the time the device info is retrieved. So for every time we get all devices or get one device in the device controller, perform the check.
- When retrieving the streams of a device that is offline, set status to online and generate a log

sebhani:
**Issue number**: #218 

**Task description**: 

 - Maintain Status field for the Device Entity
 - Generate a log when the status changes

# Additional notes

**Note to reviewers**:
The dates should be saved in UTC time zone in the DB (for lastCommunication field). I used the following website to get the UTC time: https://time.is/UTC
 
To ease your testing, I have saved the requests that I used to test the functionality when retrieving all decoders on the team's workspace in postman: 
image

1- create a device
2- create a decoder: if you set the lastCommunication to be more than 10min (in UTC time), the status field in the created device should be changed to offline (maintaining the status field is triggered in the third step)
3-  retrieve all decoders so that maintaining the status field and the creation of a log functionality is triggered


felixlapierre:
Assigning @MrJCipherButtles @FGRCL and @sharon-ho to review this

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
```diff
@@ -29,24 +31,35 @@
...
+    // maintain status field and create a log if status changed
+    Optional<DecoderDto> decoder = decoderDao.findDecoder(serialNumber);
+    if (decoder.isPresent()) {
+      List<DecoderEntity> decodersListTemp = new LinkedList<>();
+      decodersListTemp.add(decoderMapper.toDecoderEntity(decoder.get()));
+      maintainDeviceStatus.maintainStatusField(decodersListTemp);
+    }
+
+    // return decoder
+    return decoder
```
mohhef: i think its better if you:

- maintain the status 
- return the properly maintained status
- use the aspect to do the logging part "afterReturning"

mohhef, line 68: here i think you are returning the decoder that is not maintained, because you find the decoder first then you maintain the status. Id do as the other comment
- maintain the status 
- return the properly maintained status
- use the aspect to do the logging part "afterReturning"

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -31,24 +33,36 @@
...
 
   @GetMapping("/{serialNumber}")
   public ResponseEntity<EncoderDto> retrieveEncoder(@PathVariable @Valid String serialNumber) {
-    return encoderDao
-        .findEncoder(serialNumber)
+
+    // maintain status field and create a log if status changed
+    Optional<EncoderDto> encoder = encoderDao.findEncoder(serialNumber);
+    if (encoder.isPresent()) {
+      List<EncoderEntity> encodersListTemp = new LinkedList<>();
```
mohhef, line 62: can we pass just the object or is there a reason we are using a list

src/main/java/org/beanpod/switchboard/util/MaintainDeviceStatus.java
```diff
@@ -0,0 +1,93 @@
...
+      if (((devices.get(i)).getDevice().getStatus()).equalsIgnoreCase(ONLINE_STATUS)
+          && dateToBeCompared.after(devices.get(i).getLastCommunication())) {
+        // update last_communication field to offline
+        (devices.get(i).getDevice()).setStatus(OFFLINE_STATUS);
+        service.save(deviceMapper.toDeviceDto(devices.get(i).getDevice()));
+
+        // create a log
+        createLog(OFFLINE_STATUS, (devices.get(i)).getSerialNumber());
+      } else if (((devices.get(i)).getDevice().getStatus()).equalsIgnoreCase(OFFLINE_STATUS)
+          && dateToBeCompared.before(devices.get(i).getLastCommunication())) {
```
mohhef: i think devices.get(i).getdevice() is used multiple times in this code. maybe could be refactored into one line
or refactor devices.get(i) in one line

src/main/java/org/beanpod/switchboard/util/MaintainDeviceStatus.java
```diff
@@ -0,0 +1,93 @@
...
+    DecoderDto decoder = streamDto.getInputChannel().getDecoder();
+    EncoderDto encoder = streamDto.getOutputChannel().getEncoder();
+
+    // update the status field for decoder
+    if ((decoder.getDevice().getStatus()).equalsIgnoreCase(OFFLINE_STATUS)) {
+      decoder.getDevice().setStatus(ONLINE_STATUS);
+      service.save(decoder.getDevice());
+
+      // create a log
+      createLog(ONLINE_STATUS, decoder.getSerialNumber());
```
mohhef: same as above maybe its better if we keep the logging into the aspect after the function finished excuting

src/main/java/org/beanpod/switchboard/util/MaintainDeviceStatus.java
```diff
@@ -0,0 +1,93 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class MaintainDeviceStatus {
+
+  private static final DateUtil date = new DateUtil();
+  private static final String ONLINE_STATUS = "online";
+  private static final String OFFLINE_STATUS = "offline";
+  private final DeviceDaoImpl service;
```
mohhef: lets change this to deviceDao instead of service because now we have a sperate service package

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
mohhef: maybe even maintaing the status could be done in service level and not util  (i.e. retrieveDecoder in service). what do you think. if itll be used by other functions then keep it at util if not then at service maybe better

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -31,24 +33,36 @@
...
+      encodersListTemp.add(encoderMapper.toEncoderEntity(encoder.get()));
+      maintainDeviceStatus.maintainStatusField(encodersListTemp);
+    }
+
+    // return encoder
+    return encoder
         .map(ResponseEntity::ok)
```
mohhef, line 72: returned encoder is not the maintained one i think

src/main/java/org/beanpod/switchboard/entity/EncoderEntity.java
```diff
@@ -29,7 +29,7 @@
 @AllArgsConstructor
 @NoArgsConstructor
 @JsonIgnoreProperties(value = {"hibernateLazyIntializer", "handler"})
-public class EncoderEntity {
+public class EncoderEntity implements DummyInterface {
```
mohhef: nice idea!

mohhef (changes requested):

src/main/java/org/beanpod/switchboard/entity/EncoderEntity.java
ButlerJAL: I'm not sure I get the idea? Can you explain why implementing a interface class with the name DummyInterface a good idea? Is there a certain design pattern you are trying to achieve here? Also as stated in another comment, I would prefer a name that is more descriptive of the interface such as EncoderDecoderInterface possibly?

src/main/java/org/beanpod/switchboard/entity/DummyInterface.java
```diff
@@ -0,0 +1,12 @@
+package org.beanpod.switchboard.entity;
+
+import java.util.Date;
+
+public interface DummyInterface {
```
ButlerJAL: Interface name is not descriptive enough in my opinion.

src/main/java/org/beanpod/switchboard/entity/DecoderEntity.java
```diff
@@ -29,7 +29,7 @@
 @AllArgsConstructor
 @NoArgsConstructor
 @JsonIgnoreProperties(value = {"hibernateLazyIntializer", "handler"})
-public class DecoderEntity {
+public class DecoderEntity implements DummyInterface {
```
ButlerJAL: As stated in another comment, I do not get the idea of implementing an interface after the fact. Did you want it so the compiler would fail if someone were to remove said methods?

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
```diff
@@ -47,6 +48,7 @@
   @Mock private DeviceDaoImpl deviceService;
   @Mock private DecoderMapper decoderMapper;
   @Mock private StreamMapper streamMapper;
+  @Mock private MaintainDeviceStatus maintainDeviceStatus;
```
ButlerJAL, line 52: Why import and add a maintainDeviceStatus field here if it is unused?

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
ButlerJAL: I agree with @mohhef here. I think it would be simpler if the device/encoder/decoder object have some status field that gets changed at the device/encoder/decoder level, in our case the sample sender and sample receiver, which they would be responsible to notify the backend using some status change http request. Like that we could have some aspect to monitor said http request method to generate logs on the status.

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
ButlerJAL: I'm not a huge fan of the logging logic being used to maintain the status field. I think this could be ultimately refactored using aspect oriented programming. I'd suggest you look at the aspect classes done by @mohhef or consult with him as inspiration to refactor your code.

src/main/java/org/beanpod/switchboard/util/MaintainDeviceStatus.java
```diff
@@ -0,0 +1,93 @@
...
+import org.beanpod.switchboard.dto.EncoderDto;
+import org.beanpod.switchboard.dto.StreamDto;
+import org.beanpod.switchboard.dto.mapper.DeviceMapper;
+import org.beanpod.switchboard.entity.DummyInterface;
+import org.beanpod.switchboard.service.LogService;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class MaintainDeviceStatus {
```
ButlerJAL, line 19: After looking more closely at this class and considering our requirements, there's a lot of logic I feel that could be reduced/avoided if we follow how @mohhef tackled logging using AOP. @sebhani, what is your gut telling you?

src/test/java/org/beanpod/switchboard/controller/EncoderControllerTest.java
```diff
@@ -47,6 +49,7 @@
   @Mock private EncoderService encoderService;
   @Mock private EncoderMapper encoderMapper;
   @Mock private StreamMapper streamMapper;
+  @Mock private MaintainDeviceStatus maintainDeviceStatus;
```
ButlerJAL, line 53: Why import and add a maintainDeviceStatus field here if it is unused?

src/test/java/org/beanpod/switchboard/fixture/ChannelFixture.java
```diff
@@ -55,7 +56,7 @@ public static OutputChannelEntity getOutputChannelEntity() {
     return OutputChannelEntity.builder()
         .id(OUTPUT_CHANNEL_ID)
         .channel(getChannelEntity2())
-        .encoder(EncoderFixture.getEncoderEntity1())
+        .encoder(EncoderFixture.getEncoderEntity2())
```
ButlerJAL, line 59: Why the change? Was this necessary for your PR?

src/test/java/org/beanpod/switchboard/util/MaintainDeviceStatusTest.java
```diff
@@ -0,0 +1,97 @@
...
+    // TEST CASE 2: status field should be updated to online
+    // -8 minutes from now
+    date.setTime(System.currentTimeMillis() - 480000);
+
+    (encodersList.get(0)).setLastCommunication(date);
+
+    maintainDeviceStatus.maintainStatusField(encodersList);
+
+    assertEquals("online", encodersList.get(0).getDevice().getStatus());
+  }
```
ButlerJAL: Based on the comments, could this test be separated into two tests?

src/test/java/org/beanpod/switchboard/util/MaintainDeviceStatusTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+    assertEquals("online", deviceDto.getStatus(), "Status attribute is expected to be online");
+    assertEquals("online", deviceDto1.getStatus(), "Status attribute is expected to be online");
+
+    // Test Case 2: status is equal to online
+    maintainDeviceStatus.maintainStatusField(streamDto);
+
+    assertEquals("online", deviceDto.getStatus(), "Status attribute is expected to be online");
+    assertEquals("online", deviceDto1.getStatus(), "Status attribute is expected to be online");
+  }
```
ButlerJAL, line 118: Based on the comments, could this test be separated into two tests?

ButlerJAL (commented):
A general comment before I dive in the review, try to avoid noise in your PRs. There are about 38 out of 49 files that are noise mostly because you ran the formatter on the entire code base which caused newlines to be added in files unrelated to this PR. So, I would suggest that you use the formatter more specifically to the files you make changes to. Also, I am not too sure how the following file change relates to your PR: `src/test/java/jmeter/Aggregate Graph post.png`. 

By avoiding noise, this will help reviewers review your PRs more quickly and efficiently.

The remainder of my comments are inline with your code.

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
sebhani: Good suggestion! I have extracted the the logging logic out of the MaintainDeviceStatus class to an Aspect class. For maintaining the status in the service level, I decided not to proceed for two reasons: 1- It's going to require more work on this US while we have others to tackle 2- I don't think there's a huge benefit from doing such a thing.

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
sebhani: @MrJCipherButtles I have extracted the logging logic to an Aspect class.

sebhani: @mohhef We could, but then maintainStatusField will be duplicated since the logic for retrieveAllEncoders and retrieveEncoder method is the same exact logic.

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
sebhani: Good catch! I've updated the retrieveEncoder/retrieveDecoder controllers to return the most updated object.

src/main/java/org/beanpod/switchboard/util/MaintainDeviceStatus.java
sebhani: Good suggestion. I've updated it.

src/main/java/org/beanpod/switchboard/util/MaintainDeviceStatus.java
sebhani: I've updated it

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
sebhani: Good catch! I've changed the code to return the most updated object.

src/main/java/org/beanpod/switchboard/entity/EncoderEntity.java
sebhani: @MrJCipherButtles this interface serves a single purpose which is avoiding duplicating maintainStatusField method. If DecoderEntity and EncoderEntity classes don't implement a common interface, then we would have to write one method for each although the logic is exactly the same. I've added a comment on the top of the interface to give clear idea what this interface is for.

src/main/java/org/beanpod/switchboard/entity/DummyInterface.java
sebhani: Good name suggestion! I've followed that.

src/main/java/org/beanpod/switchboard/entity/DecoderEntity.java
sebhani: I explained the reason be hind the DummyInterface in an earlier comment: https://github.com/bean-pod/switchboard/pull/304#discussion_r574583109

src/test/java/org/beanpod/switchboard/controller/DecoderControllerTest.java
sebhani: That's true! Now it's being used.

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
sebhani: I have extracted the logging logic to an Aspect class.

src/test/java/org/beanpod/switchboard/controller/EncoderControllerTest.java
sebhani: It's being used right now.

src/test/java/org/beanpod/switchboard/fixture/ChannelFixture.java
sebhani: Yes it's. Previously, both input and output channels fixtures had the same device which makes no sense. Therefore, I changed it so that the output channel has a different device than the input channel.

src/test/java/org/beanpod/switchboard/util/MaintainDeviceStatusTest.java
sebhani: I don't think it's necessary to divide these into separate tests. These cases are testing the same exact functionality, but with different timing setup.

src/test/java/org/beanpod/switchboard/util/MaintainDeviceStatusTest.java
sebhani: Both test cases are testing the same exact functionality. I'm just making sure that the if statement is working properly.

src/main/java/org/beanpod/switchboard/util/MaintainDeviceStatus.java
sebhani: I've extracted the logging functionality to an Aspect class.

ButlerJAL (commented):
My general thoughts is that the code seems a bit much for ultimately logging the changes of an object's field.

For example, the conditional logic in MaintainDeviceStatus is mostly for comparing timestamps and what not. If we are worried about stale data affecting a status' true value, maybe a better place to deal with that would be directly in the database using a MySQL scheduled event that directly modifies the status to offline when last communication is greater than 10 minutes. Functions related purely to data is better left to be done at the database level. This also lessens the responsibility of the backend service.

I guess I'm trying to wrap my head around what value are we really getting with all this new code we'll ultimately need to maintain in the backend. I'd like to hear from @mohhef or anyone, if they see what I am seeing in that the code seems a bit much for what it is.

I could be wrong, so, maybe if I could be told as to why we need a big MaintainDeviceStatus class with generics, conditional logic, having objects with status store temporary lists in memory... Because I see it more sensible to have an Enum with status values which we could monitor any status mutators with an aspect. And, regarding automatically updating a status to offline for inactivity this can be done directly in the database.

I hope this comment does not brush off as me bashing on your code. I mostly wanted to share the alternative that makes more sense to me, and I believe may be more appropriate. Possibly my alternative may not be clear short like this in a comment, so, maybe all I am saying could be considered for a future refactor.

I'm not going to request any changes per say because ultimately what I am saying is that I'd prefer to go about the solution differently; however, the solution we implement is a team call.

Overall, I think your solution works albeit I think that it may possibly prove to be hard to work with going forth which is my two cents, but if the consensus is that we like this code, and the complexity is acceptable such that a refactor (alternative solution) before merging is unnecessary, I would not object to follow up this review with an approval.

sebhani:
> My general thoughts is that the code seems a bit much for ultimately logging the changes of an object's field.
> 
> For example, the conditional logic in MaintainDeviceStatus is mostly for comparing timestamps and what not. If we are worried about stale data affecting a status' true value, maybe a better place to deal with that would be directly in the database using a MySQL scheduled event that directly modifies the status to offline when last communication is greater than 10 minutes. Functions related purely to data is better left to be done at the database level. This also lessens the responsibility of the backend service.
> 
> I guess I'm trying to wrap my head around what value are we really getting with all this new code we'll ultimately need to maintain in the backend. I'd like to hear from @mohhef or anyone, if they see what I am seeing in that the code seems a bit much for what it is.
> 
> I could be wrong, so, maybe if I could be told as to why we need a big MaintainDeviceStatus class with generics, conditional logic, having objects with status store temporary lists in memory... Because I see it more sensible to have an Enum with status values which we could monitor any status mutators with an aspect. And, regarding automatically updating a status to offline for inactivity this can be done directly in the database.
> 
> I hope this comment does not brush off as me bashing on your code. I mostly wanted to share the alternative that makes more sense to me, and I believe may be more appropriate. Possibly my alternative may not be clear short like this in a comment, so, maybe all I am saying could be considered for a future refactor.
> 
> I'm not going to request any changes per say because ultimately what I am saying is that I'd prefer to go about the solution differently; however, the solution we implement is a team call.
> 
> Overall, I think your solution works albeit I think that it may possibly prove to be hard to work with going forth which is my two cents, but if the consensus is that we like this code, and the complexity is acceptable such that a refactor (alternative solution) before merging is unnecessary, I would not object to follow up this review with an approval.

I thought about using MySQL scheduler to run a script to update the status field every 10min. However, I am always against modifying data directly in the DB because of the logic that can be missed. You can read about this notion online.
I got this idea from my first internship where we used to maintain data of specific fields using plugins (for D365 CRM) and not writing a PS/SSIS script to run every X minutes although it was way easier.

mohhef:
> t it may possibly prove to be hard to work with going forth which is my two cents, but if the consensus is that we like this code, and the complexity is acceptable such that a refactor (alternative solution) before merging is unnecessary, I would not object to follow up this review with an approva

@MrJCipherButtles I see what you see from the point where this could be included in the db side instead of our logic side. I agree it could be a future refactor. But i also think using something else would still require maintaining, and probably an overhead of learning.

If we would consider this refactor, we can use sql triggers, basically its an action that is triggered on an event. So the trigger would basically do @sebhani  logic when we are getting encoder/decoder.
However, when it comes down to triggers, it is hard for future maintainers to understand and sometimes even actually know they exist (maintainers here i mean the actual PO,PM) because the syntax varies quite a bit from different sql languages, and since we are using JPA, that trigger could be over triggered, in the sense of, the trigger should be triggered when we try to make a **select**  query on a "get" request but JPA actually uses **select** in many of its queries, not just the "get" requests, which actually could be a good or a bad thing (not too sure).

note: ill try to rereview your pr tomorrow morning

mohhef:
@sebhani @MrJCipherButtles 
I actually found this which is alike triggers using JPA: https://docs.jboss.org/hibernate/stable/entitymanager/reference/en/html/listeners.html
https://www.baeldung.com/jpa-entity-lifecycle-events
could be really interesting for future reference

src/main/java/org/beanpod/switchboard/aop/MaintainDeviceStatusAspect.java
```diff
@@ -0,0 +1,46 @@
...
+      value =
+          "execution(* org.beanpod.switchboard.util.MaintainDeviceStatus"
+              + ".maintainStatusField(java.util.List))",
+      returning = "updatedDevices")
+  public void createLogEncoderAndDecoder(List<DeviceEntity> updatedDevices) {
+    createLogHelper(updatedDevices);
+  }
+
+  // helper method to create logs
+  private void createLogHelper(List<DeviceEntity> updatedDevices) {
```
felixlapierre: I'd rename this to `createLog`, with the current name i initally assumed this method created someting called a `LogHelper`

src/main/java/org/beanpod/switchboard/util/MaintainDeviceStatus.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    return updatedDevices;
+  }
+
+  private Date getDateToBeCompared() {
+    // subtract 10minutes from the current date
+    Date dateToBeCompared = date.getCurrentDate();
+    TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
+    dateToBeCompared.setTime(System.currentTimeMillis() - 600000);
```
felixlapierre, line 91: We should make the duration (10 minutes) configurable through spring profiles

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
```diff
@@ -29,24 +32,38 @@
...
   public ResponseEntity<DecoderDto> retrieveDecoder(@PathVariable @Valid String serialNumber) {
-    return decoderDao
-        .findDecoder(serialNumber)
+    // maintain status field and create a log if status changed
+    Optional<DecoderDto> decoder = decoderDao.findDecoder(serialNumber);
+    if (decoder.isPresent()) {
+      List<DecoderEntity> decodersListTemp = new LinkedList<>();
+      decodersListTemp.add(decoderMapper.toDecoderEntity(decoder.get()));
+      DeviceEntity updatedDevice =
+          maintainDeviceStatus.maintainStatusField(decodersListTemp).get(0);
```
felixlapierre, line 62: To simplify the logic here, i would add a method to `MaintainDeviceStatus` that handles one device e.g.

```
public <T extends DecoderEncoderInterface> DeviceEntity maintainStatusField(T device) {
    return maintainStatusField(List.of(device)).get(0);
}
```
then in this class and EncoderController we can do
```
DecoderEntity decoderEntity = decoderMapper.toDecoderEntity(decoder.get());
DeviceEntity updatedDevice = maintainDeviceStatus.maintainStatusField(decoderEntity);
```

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -31,24 +34,39 @@
...
+
+    // maintain status field and create a log if status changed
+    Optional<EncoderDto> encoder = encoderDao.findEncoder(serialNumber);
+    if (encoder.isPresent()) {
+      List<EncoderEntity> encodersListTemp = new LinkedList<>();
+      encodersListTemp.add(encoderMapper.toEncoderEntity(encoder.get()));
+      DeviceEntity updatedDevice =
+          maintainDeviceStatus.maintainStatusField(encodersListTemp).get(0);
+      // update the retrieved decoder object
+      encoder.get().getDevice().setStatus(updatedDevice.getStatus());
```
mohhef, line 67: should this exist in the service

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
```diff
@@ -29,24 +32,38 @@
...
+      // update the retrieved decoder object
+      decoder.get().getDevice().setStatus(updatedDevice.getStatus());
```
mohhef, line 64: should this be in the service

mohhef (dismissed):
i had another small comment and it would be great if you can cover the edge cases not covered 
https://sonarcloud.io/component_measures?id=bean-pod_switchboard&pullRequest=304&metric=new_coverage&view=list
Otherwise LGTM!

ButlerJAL:
@mohhef what I was suggesting is a [scheduled event](https://www.mysqltutorial.org/mysql-triggers/working-mysql-scheduled-event/). It’s not exactly the same as a trigger although some use the term “temporal trigger” to describe it. A scheduled event is not something you would do in JPA. The addition of a scheduled event is done using SQL code maybe five or so lines through a database console.

Maintainability of a scheduled event is not that difficult considering it’s only a few lines. Learnability, although considered a lesser known more advance feature of SQL, is not really time-consuming.

@sebhani did bring up good points as to why it’s not typically a recommended practice to use said feature; however, it seems to me the goal here is to not have expired data being used by our service. The distinction being that our service is dependent on the data received. In this case, something like a SQL scheduled event seems fitting for the application.

You look at the code as it is now, our service is calling the database to get encoders and decoders. Our service filters through to find all encoders and decoders with expired timestamps to then update their statuses to offline, and every time an expired timestamps is found we send back an update query to the database. Hey it works, but it seems like quite a roundabout and inefficient way of doing something we could otherwise do by leveraging the scheduled event feature of SQL which avoids our service to ever worry about getting expired data in the first place. Which in other words translates to there actually being a lesser worry for maintainers on the service side.

However, like @sebhani said, I am aware that scheduled events can cause side effects. I’m just not convinced there is one for our application. If there is well then I’d be more inclined to fully accept this solution such that the validation of data is done in the service, else it’s not my favorite solution albeit a solution. Especially that I’m looking a little further in that we’ll most likely need to add more conditional logic to this code to possibly accommodate other statuses then online and offline.

mohhef (dismissed):
Ill approve this for now as it seems to properly do what its supposed to do, we can discuss the other ideas next meeting

mohhef:
@MrJCipherButtles  What i meant to say is there are triggers that are done on the db side, it looks something like this :https://dev.mysql.com/doc/refman/8.0/en/trigger-syntax.html
It has nothing to do with JPA, but it turns out that you can code the trigger from the JPA side (i.e. that means that we dont need to maintain the trigger separately, it will be created with the creation of all our other tables)

its like a scheduled event but its an event driven event (when something happens->do something instead of doing something every x minutes or so)

felixlapierre (commented):
Just a few lil suggestions, otherwise lgtm

src/main/java/org/beanpod/switchboard/controller/DecoderController.java
felixlapierre: We don't have this method in the service yet, though we have a refactoring task in zenhub to move it, so we'll move this code when we do that refactoring task

felixlapierre (approved):
My fixes PR was merged so LGTM

mohhef (approved):
LGTM

PR #305, opened 2021-02-08
[#275] Create new account

### LINKED ISSUE #275: Endpoint to create a new account
Create an endpoint to create a new account in the system.

This endpoint should only be called by a user who is an admin user, and should only be able to create non-admin users.

mohhef:
**Issue number**: 275

**Task description**: 

- Endpoint for creating an account with email, name and password is created
- Commeneted out code is related with spring security config, @FGRCL will need this code, enabiling that code will enable spring security on the whole system
- Enabling spring security for the whole system is needed but only when we can sign in
- Enabling spring security will fail all the integration tests, a work around should be found when sign in is created in francois task

The security issue needs to be approved by the reviewer to verify that it is not an issue, I have looked at it myself and its not an issue
1st issue : it just returns if the user is admin or regular


felixlapierre:
Assigning this to @sebhani @FGRCL and @pobudzey for review

src/main/java/org/beanpod/switchboard/entity/UserEntity.java
```diff
@@ -0,0 +1,31 @@
...
+@NoArgsConstructor
+@AllArgsConstructor
+@Entity(name = "User")
+public class UserEntity {
+
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  private Long id;
+
+  @NotNull private String userName;
```
FGRCL: I think username is a single word in this context.

src/main/java/org/beanpod/switchboard/entity/MyUserDetails.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+import javax.validation.constraints.NotNull;
+import lombok.Builder;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.authority.SimpleGrantedAuthority;
+import org.springframework.security.core.userdetails.UserDetails;
+
+public class MyUserDetails implements UserDetails {
```
FGRCL: What is the difference between this class and `UserEntity`, they seem to have some common fields. I would also call this class `SwitchboardUserDetails` instead.

src/main/java/org/beanpod/switchboard/repository/UserRepository.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.Optional;
+import org.beanpod.switchboard.entity.UserEntity;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface UserRepository extends JpaRepository<UserEntity, Long> {
+
+  Optional<UserEntity> findByUserName(String email);
```
FGRCL: This is small, but I think it should be `username` now instead of `email`

src/main/java/org/beanpod/switchboard/controller/UserController.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Controller
+@AllArgsConstructor
+public class UserController implements UserApi {
+
+  private final UserDaoImpl userDao;
+
+  @Override
+  public ResponseEntity<String> signUp(UserModel user) {
+    userDao.save(user);
```
FGRCL: Your controller should call your `UserService` instead.

src/main/java/org/beanpod/switchboard/dao/UserDaoImpl.java
```diff
@@ -0,0 +1,20 @@
...
+
+@Component
+@RequiredArgsConstructor
+public class UserDaoImpl {
+
+  private final UserService userService;
+  private final UserMapper userMapper;
+
+  public UserDto save(UserModel userModel) {
+    return userMapper.toUserDto(userService.signUpUser(userMapper.toUserEntity(userModel)));
```
FGRCL: Your `UserDao` should be the one interacting with the Repository to save your `UserEntity`

src/main/java/org/beanpod/switchboard/dto/mapper/UserMapper.java
```diff
@@ -0,0 +1,21 @@
...
+import org.beanpod.switchboard.entity.UserEntity;
+import org.mapstruct.Mapper;
+import org.openapitools.model.UserModel;
+
+@Mapper(componentModel = "spring")
+public interface UserMapper {
+
+  UserDto toUserDto(UserEntity userEntity);
+
+  DeviceDto toUserDto(UserModel userModel);
```
FGRCL: There's a typo here, the return type should be `UserDto`.

src/main/java/org/beanpod/switchboard/service/UserService.java
```diff
@@ -0,0 +1,39 @@
...
+@Service
+@RequiredArgsConstructor
+public class UserService implements UserDetailsService {
+
+  private final UserRepository userRepository;
+  private final BCryptPasswordEncoder bCryptPasswordEncoder;
+
+  @Override
+  public UserDetails loadUserByUsername(String email) {
+    final Optional<UserEntity> optionalUser = userRepository.findByUserName(email);
```
FGRCL: As I've mentioned, this database access logic should be moved to the `UserDao`

src/main/java/org/beanpod/switchboard/service/UserService.java
```diff
@@ -0,0 +1,39 @@
...
+      throw new UsernameNotFoundException(
+          MessageFormat.format("User with username {0} cannot be found.", email));
+    }
+  }
+
+  public UserEntity signUpUser(UserEntity user) {
+    String password = user.getPassword();
+    final String encryptedPassword = bCryptPasswordEncoder.encode(password);
+    user.setPassword(encryptedPassword);
+    return userRepository.save(user);
```
FGRCL: Same thing about the database access logic. I think the password encryption does belong in the service thought so that's good.

src/test/java/org/beanpod/switchboard/service/UserServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @BeforeEach
+  void setupUserFixture() {
+    userEntity1 = UserFixture.getUserEntity();
+  }
+
+  @BeforeEach
+  void setup() {
+    MockitoAnnotations.initMocks(this);
+  }
```
FGRCL: any reason why we're using two separate setup methods here?

src/test/java/org/beanpod/switchboard/service/UserServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.jupiter.api.Test;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.springframework.security.core.userdetails.UsernameNotFoundException;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+
+class UserServiceTest {
+
+  @Mock private static UserEntity userEntity;
```
FGRCL: I think this variable is unused in the tests

FGRCL (commented):
Looks almost good to go, I have a couple of comments/questions

src/main/java/org/beanpod/switchboard/entity/UserEntity.java
mohhef: fixed in next push

src/main/java/org/beanpod/switchboard/entity/MyUserDetails.java
mohhef: loadUserByUsername method in the service has to return a UserDetails type, and the UserDetails service has to be implemented as a result we have both. We can actually have them all in one class but i read that is not recommended

src/main/java/org/beanpod/switchboard/dto/mapper/UserMapper.java
mohhef: fixed in next push

src/test/java/org/beanpod/switchboard/service/UserServiceTest.java
mohhef: i have always separated the Mockito one from the variables setup, not sure why but ill just merge them to one

src/test/java/org/beanpod/switchboard/service/UserServiceTest.java
mohhef: used in line 56

src/main/java/org/beanpod/switchboard/repository/UserRepository.java
mohhef: fixed in next push

src/main/java/org/beanpod/switchboard/service/UserService.java
FGRCL: While testing I found that I was able to create the same user multiple times. We should add a check here to make sure it doesn't happen

src/main/java/org/beanpod/switchboard/controller/UserController.java
mohhef: fixed, i actually messed that up haha

src/main/java/org/beanpod/switchboard/dao/UserDaoImpl.java
mohhef: fixed

src/main/java/org/beanpod/switchboard/service/UserService.java
mohhef: fixed

src/main/java/org/beanpod/switchboard/service/UserService.java
mohhef: fixed

src/test/java/org/beanpod/switchboard/service/UserServiceTest.java
mohhef: fixed, removed

src/main/java/org/beanpod/switchboard/dao/UserDaoImpl.java
```diff
@@ -0,0 +1,36 @@
...
+public class UserDaoImpl {
+
+  private final UserRepository userRepository;
+  private final UserMapper userMapper;
+
+  public UserDto save(UserModel userModel) {
+    return userMapper.toUserDto(userRepository.save(userMapper.toUserEntity(userModel)));
+  }
+
+  public UserDetails loadUserByUsername(String email) {
```
FGRCL: We should change this param name to `username`

src/main/java/org/beanpod/switchboard/service/UserService.java
```diff
@@ -0,0 +1,30 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class UserService implements UserDetailsService {
+
+  private final UserDaoImpl userDao;
+  private final BCryptPasswordEncoder bCryptPasswordEncoder;
+
+  @Override
+  public UserDetails loadUserByUsername(String email) {
```
FGRCL: this parameter name could also be changed to `username`

FGRCL (dismissed):
Damn you reviewed those comments fast! I just have some rename suggestions that I missed the first time, so I'm not going to let that block the PR. LGTM

src/main/java/org/beanpod/switchboard/service/UserService.java
mohhef: ah i keep missing these

src/main/java/org/beanpod/switchboard/service/UserService.java
mohhef: good idea, done

mohhef: feel free @FGRCL  to rereview

src/main/java/org/beanpod/switchboard/dao/UserDaoImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    return userMapper.toUserDto(userRepository.save(userMapper.toUserEntity(userModel)));
+  }
+
+  public UserDetails loadUserByUsername(String username) {
+    final Optional<UserEntity> optionalUser = userRepository.findByUsername(username);
+    if (optionalUser.isPresent()) {
+      return new SwitchBoardUserDetails(optionalUser.get());
+    } else {
+      throw new UsernameNotFoundException(
+          MessageFormat.format("User with username {0} cannot be found.", username));
```
ButlerJAL: Using a more generic message such as "bad credentials" will get rid of one of your security hotspots.

ButlerJAL (changes requested):

src/main/java/org/beanpod/switchboard/dao/UserDaoImpl.java
mohhef: sonar still says its a security issue after fixing, it actually didnt use to in the more previous commits although i had the same old logic. I dont know if im missing something

src/main/java/org/beanpod/switchboard/config/WebConfig.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+
+@Configuration
+public class WebConfig {
+
+  @Bean
+  public BCryptPasswordEncoder bCryptPasswordEncoder() {
```
ButlerJAL: To reduce the code smell related to `Method name 'bCryptPasswordEncoder' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9_]*$'`. I'd suggest refactoring `bCryptPasswordEncoder` --> `bcryptPasswordEncoder` or `passwordEncoder`. Any name such that the second character is a lowercase letter, or number would be compliant.

src/main/java/org/beanpod/switchboard/config/WebSecurityConfig.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+
+@Configuration
+@AllArgsConstructor
+public class WebSecurityConfig extends WebSecurityConfigurerAdapter {
+
+  private final UserService userService;
+
+  private final BCryptPasswordEncoder bCryptPasswordEncoder;
```
ButlerJAL: To reduce the code smell related to `Member name 'bCryptPasswordEncoder' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'`. I'd suggest refactoring `bCryptPasswordEncoder` --> `bcryptPasswordEncoder` or `passwordEncoder`. Any name such that the second character is a lowercase letter, or number would be compliant.

src/main/java/org/beanpod/switchboard/service/UserService.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class UserService implements UserDetailsService {
+
+  private final UserDaoImpl userDao;
+  private final UserRepository userRepository;
+  private final BCryptPasswordEncoder bCryptPasswordEncoder;
```
ButlerJAL: To reduce the code smell related to `Member name 'bCryptPasswordEncoder' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'`. I'd suggest refactoring `bCryptPasswordEncoder` --> `bcryptPasswordEncoder` or `passwordEncoder`. Any name such that the second character is a lowercase letter, or number would be compliant.

ButlerJAL (changes requested):
I've investigated that the security hotspots are in fact valid, anything security related will be flagged for review by Sonar. Security hotspots don't necessarily mean that something is wrong, but that we should review whether or not they are applicable and of concern to us.

The last commit lessened the security vulnerability related to an attacker's ability to guess existing usernames because the UsernameNotFoundException() is thrown without including the username that triggered the exception. But, without getting into details, it's not much for a motivated attacker to still somewhat extract information based on the thrown exception.

It is possible in Spring Security to set HideUserNotFoundExceptions to true in our DaoAuthenticationProvider which as the name would imply hides UsernameNotFoundException(). In our case, we should be able to do this in configureGlobal() of WebSecurityConfig.java

Obviously, hiding thrown exceptions could prove to be a debugging nightmare; thus, we need to ensure that we throw a BadCredentialsException() when UsernameNotFoundException() is caught.

Looking more thoroughly at this PR, from my understanding we are mostly concerned with signing up users using our database for persistence which looks good, especially thanks to @FGRCL reviews. As there are not many security concerns related to signing up, I think it'll be important for @FGRCL who is working on both the spring security configuration where he is defining roles/authorities and the signing in logic to worry about roles/authorities and thrown authentication exceptions that are more specific to signing in.

With all this said, when it comes to security, the goal is to make any potential attacker's job as difficult as possible, and one way you do that is to give the least amount of actionable information as possible. Practically, our security hotspots are of low priority because there's only so much an attacker could exploit here, so, as long as we are aware I wouldn't overthink fixing them if we wanted to ignore.

@mohhef I suggested changes to reduce your code smells; otherwise, this PR LGTM for approval 👍. I've reviewed the security hotspots as safe for this PR. Now the checks pass for this PR. We may want to re-review that the security hotspots remain safe once @FGRCL submits his follow-up PR.

src/main/java/org/beanpod/switchboard/config/WebConfig.java
mohhef: fixed

src/main/java/org/beanpod/switchboard/config/WebSecurityConfig.java
mohhef: fixed

src/main/java/org/beanpod/switchboard/service/UserService.java
mohhef: fixed

ButlerJAL (approved):

src/main/java/org/beanpod/switchboard/dao/UserDaoImpl.java
ButlerJAL: [resolved](https://github.com/bean-pod/switchboard/pull/305#pullrequestreview-590786489)

FGRCL (approved):
Reapproving

PR #308, opened 2021-02-09
[#273] Add login page to frontend

### LINKED ISSUE #273: Log-in page
Create a log in page, that takes a username and password (or whatever our choice of authentication method requires), has a login button.

* If login is successful, persist the session and redirect to the homepage
* If login fails, display an error and redirect to / remain on the log in page.

Calls the endpoint created in #272

For convenience, we should make sure that you don't need to log back in to the frontend constantly. The session should persist if you restart the frontend.

felixlapierre:
# General info

**Issue number**: #273

**Task description**: 

 - Add a login page for the user to log in to the service
 - Create an error popup to display when login errors occur

**What is not included in this PR:**

- Right now the login page works and is tested, but doesn't actually connect to the backend. I'll add that in a future PR once the backend task for this story is done.

**Screenshots**
 
image

image

# Testing

**Test coverage**:

Unit tested the new components (80.6% coverage). No tests for LoginPage.jsx since i was struggling to make a test that only tested that component.


felixlapierre:
Assigning @gdzooks @arejayelle and @mohhef to review this

gdzooks (changes requested):
alright so the only changes i'll request are 
- fixing the code smells for the prop validation if you can! and
- pulling from master to get the delete button in correct folder after my last PR

other than that
- it might be nice if you could get the login page to show first as default, but i don't really mind if we don't have it in this PR since the backend isn't done yet and we'll have to update things in a big way once it is
- as far as testing the login page goes - you could do the very short basic test that a lot of other frontend elements like it have: i.e. check that it has the right child elements on the page (loginconsole and the dialog)

frontend/src/login/LoginConsole.jsx
```diff
@@ -0,0 +1,72 @@
...
+    width: "100%",
+    marginTop: theme.spacing(1)
+  },
+  submit: {
+    margin: theme.spacing(3, 0, 2)
+  }
+}));
+
+export default function LoginConsole(props) {
+  const { handleSubmit } = props;
```
sharon-ho, line 22: Missing props validation for `handleSubmit`

frontend/src/login/LoginFailedDialog.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+const useStyles = makeStyles((theme) => ({
+  title: {
+    color: "#ef5350"
+  }
+}));
+
+export default function LoginFailedDialog(props) {
+  const classes = useStyles();
+  const { open, setOpen, message } = props;
```
sharon-ho, line 21: Missing props validation for `message`

frontend/src/login/LoginFailedDialog.jsx
```diff
@@ -0,0 +1,51 @@
...
+  DialogContentText,
+  DialogTitle,
+  makeStyles
+} from "@material-ui/core";
+
+const useStyles = makeStyles((theme) => ({
+  title: {
+    color: "#ef5350"
+  }
+}));
```
sharon-ho: Why did you choose to do `useStyles` instead of defining and using classes in `index.css`?

frontend/src/login/LoginFailedDialog.jsx
```diff
@@ -0,0 +1,51 @@
...
+  Button,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogContentText,
+  DialogTitle,
+  makeStyles
+} from "@material-ui/core";
+
+const useStyles = makeStyles((theme) => ({
```
sharon-ho: There's a code smell here for `'theme' is defined but never used (no-unused-vars)`, can this be resolved by removing the keyword `theme`?

frontend/src/login/LoginConsole.jsx
```diff
@@ -0,0 +1,72 @@
...
+    alignItems: "center"
+  },
+  form: {
+    width: "100%",
+    marginTop: theme.spacing(1)
+  },
+  submit: {
+    margin: theme.spacing(3, 0, 2)
+  }
+}));
```
sharon-ho, line 19: Why did you choose to do `useStyles` instead of defining and using classes in `index.css`?

frontend/src/login/LoginPage.jsx
```diff
@@ -0,0 +1,31 @@
...
+  const handleSubmit = (username, password) => {
+    AuthenticationApi.logIn({ username, password }).catch((error) => {
+      setDialogMessage(error.message);
+      setDialogOpen(true);
+    });
+  };
+
+  return (
+    <Container>
+      <Box className="flexContents headerAreaUnderline">
```
sharon-ho, line 20: I would suggest from a design standpoint to have only the title centered as `Login` placed right above the `LoginConsole`  to remove the redundancy of having Switchboard mentioned twice in the same page (once in the header and another time on the page itself)

### Result:
| Original   | Updated|
| image | image |


### Code:
- LoginPage.jsx
image

- index.css
image

- LoginConsole.jsx
  - Remove `marginTop: theme.spacing(8)` under `paper`
  image

sharon-ho (changes requested):
Minor suggestions about code smells and some styling/implementation choices, otherwise good work!

frontend/src/api/AuthenticationApi.js
```diff
@@ -0,0 +1,22 @@
+import axios from "axios";
+
+export const unknownErrorMessage =
+  "An unknown error occurred. Please try again later";
```
sharon-ho: Being nitpicky but adding a period after later would be nice

frontend/src/login/LoginFailedDialog.jsx
felixlapierre: Honestly, I was following an example on the material ui website. My arguments in favor of keeping it this way are
- `index.css` is already huge, making it hard to find stuff in right now
- having the CSS here makes it easy to reference directly from the code it affects and avoids having side effects on other pages

I would be okay with moving this stuff into a CSS file instead, but i would want it to be specific to the login page

frontend/src/login/LoginConsole.jsx
felixlapierre: See my other comment

frontend/src/login/LoginPage.jsx
felixlapierre: Before starting the story, Jun created a wireframe of the page (see the comment on #273). so I'll stick with that unless Jun decides to change the design.

felixlapierre:
@gdzooks 
>     * fixing the code smells for the prop validation if you can!
Done!

>     * pulling from master to get the delete button in correct folder after my last PR
Done!
>     * it might be nice if you could get the login page to show first as default
Plan is to do this once it's working with the backend and we integrate the session with the routing. It will always redirect to log in unless you are logged in. This will break system tests tho but we'll cross that bridge when we get there. Maybe have an environment variable to disable it in test environments.
>     * as far as testing the login page goes - you could do the very short basic test 
Done!

frontend/src/login/LoginPage.jsx
sharon-ho: @arejayelle What do you think about the above?

frontend/src/login/LoginFailedDialog.jsx
sharon-ho: I think it would be better to have it in `index.css`. The reason it's like that in the Material UI documentation is to just show the user what styles are applied for that example specifically. You could add a comment in the CSS file to specify that they are for login. I think it's better if all styles can be found centralized in one file versus scattered around the project since some might be reused in the future for other components.

frontend/src/login/LoginPage.jsx
jun-loke: @sharon-ho I would have preferred that you made a wireframe/lofi mockup as a proof of concept instead of coding it yourself.
I agree with the change of page name, however, the removal of the line and other title stylings I disagree with. This has otherwise been consistent throughout the rest of the application. 
I can see the argument for centering it because the login card is centered and that causes unnecessary distance betweent the two elements on the page, however, I would rather move the card over to match the title and add a sort of splash image to the right instead

jun-loke: that being said, I haven't had the time to iterate over the design of this page much. So there's definitely room for change.

sharon-ho: @arejayelle Okay thanks for explaining your reasoning! In that case I'll leave it up to @felixlapierre to decide what he would like to do!

gdzooks (dismissed):
all my comments addressed! tests look delicious 😋 :tada:

frontend/src/login/LoginPage.jsx
```diff
@@ -0,0 +1,31 @@
...
+        <div className="title">Welcome to Switchboard</div>
```
jun-loke: Title to be changed, as discussed
```suggestion
        <div className="title">Login</div>
```

frontend/src/login/__tests__/LoginPage.test.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+  it("Renders with the correct elements", () => {
+    expect(wrapper.find(".title")).toHaveLength(1);
+    expect(wrapper.find("input")).toHaveLength(2);
+    expect(wrapper.find("label")).toHaveLength(2);
+    expect(wrapper.find("label").at(0).text()).toEqual("Username *");
+    expect(wrapper.find("label").at(1).text()).toEqual("Password *");
+    expect(wrapper.find("button")).toHaveLength(1);
+    expect(wrapper.find("button span.MuiButton-label").text()).toEqual(
+      "Log In"
```
jun-loke: would prefer to have more granularity for the writing of this test eg:

- it( renders with the correct elements)
  - one title
  - two text boxes
  - etc.
 
 Which can be done by nesting `it()` for better readability, but it's not major

jun-loke (changes requested):
Small change to be added as discussed and just a note of preference on the testing format. Nothing major. Will approve once you change the title

sharon-ho (dismissed):
Looks good!

gdzooks (approved):
lgtm :,3

jun-loke (approved):
All my comments have been addressed. Good job

PR #310, opened 2021-02-10
[#309] Fixed most frontend code smells

### LINKED ISSUE #309: Reduce code smells

felixlapierre:
# General info

**Issue number**: #309

**Task description**: 

- Fix frontend code smells. There are now 4 code smells remaining in the frontend.
- Reduce noise in frontend test run output, which currently has a lot of error messages printed even when all tests passed

## Changelog

- Fixed a bunch of issues in tests where we did `expect(...).toBeNull` or `expect(...).toBeFalsy` which does not actually call the assertion. I fixed these to do the proper assertion, and fixed shortcomings in the tests that were revealed by this change.
- Removed a bunch of console references. We had a lot of tests that were testing props validation, which required us to assert that console.error was being called. This was annoying not only for the code smell but for the output that was generated during the test run (there was lots of error logging despite all the tests passing and it made it confusing when looking for test failure issues). In the end i decided to remove these tests, after doing some research and seeing that most people recommend not testing prop validation.
- Removed some unused variables
- Changed components with no state into functional components
- Added display names to unnamed components. This code smell happens when we create a component in an anonymous function and return it. I fixed this issue by naming those functions, which makes the returned component take on the name of the function. The display name is used in error logging in the browser.
- Removed prop spreading, where possible. Prop spreading remains in a couple of places where I couldn't get rid of it.
- Removed usage of unary ++ operator by using the index argument of the map function instead.
- Moved functions that did not reference "this" out of the class / removed them entirely.
- Added `jest-canvas-mock` as a dependency and imported it in `setupTests.js`. This gets rid of a ton of error logs that occur when an HTML canvas is used in Jest. I didn't think we were using an html canvas, but i guess one of the material ui components uses it, since at least 5 of our tests had this massive error log.
- Fixed an uncaught promise in the StreamApi tests, one uncaught promise error remains but that will have to be fixed when we revisit those tests (see below)

## What didn't get fixed
- The code smells in the StreamApi tests are more of a symptom that those tests are not asserting the returned values are correct. I am leaving those smells as a reminder/todo to revisit those tests again sometime, since i didn't have time to fix the tests myself.
- Some of the props spreading issues in index.jsx which arise as a result of using the router and which after a quick google seem to have no solution other than telling eslint to ignore them.
# Testing

**Test coverage**: Should be the same; although some tests were modified or removed, this should not decrease coverage.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -5275,6 +5275,12 @@
       "resolved": "https://registry.npmjs.org/cssesc/-/cssesc-3.0.0.tgz",
       "integrity": "sha512-/Tb/JcjK111nNScGob5MNtsntNM1aCNUDipB/TkwZFhyDrrE47SOx/18wF2bbjgc3ZzCSKW1T5nt5EbFoAz/Vg=="
     },
+    "cssfontparser": {
+      "version": "1.2.1",
+      "resolved": "https://registry.npmjs.org/cssfontparser/-/cssfontparser-1.2.1.tgz",
+      "integrity": "sha1-9AIvyPlwDGgCnVQghK+69CWj8+M=",
+      "dev": true
+    },
     "cssnano": {
       "version": "4.1.10",
       "resolved": "https://registry.npmjs.org/cssnano/-/cssnano-4.1.10.tgz",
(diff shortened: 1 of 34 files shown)


ButlerJAL (approved):
Wow! This looks really great 👍

sharon-ho (approved):
LGTM! Really glad you took the initiative to fix most of them in this PR!

Issue #278, opened 2021-01-28
Require authentication on all service endpoints


felixlapierre:
Add authentication to all our device- and stream-related controllers

* Device controller
* Encoder controller
* Decoder controller
* Streams controller
* Channel controller

Issue #279, opened 2021-01-28
Authenticate to device/stream endpoints in frontend


felixlapierre:
After #288 all endpoints for devices/streams/etc will require authentication. This story represents the work of adding authentication to all requests we make to the backend.

PR #314, opened 2021-02-19
[#272] Add endpoint for authentication

### LINKED ISSUE #272: Endpoint to log in / authenticate
Create an endpoint to log in using the authentication method selected in #221

We need an endpoint for the frontend to call that will return a token to be used in the frontend for authentication. We also need the sample apps to be able to authenticate to indicate which user they belong to. Either we need one endpoint that can handle both.

We should also create a "default user" that is a superuser when the server starts on a fresh database. The credentials must be specified through the environment variables / Spring active profile, so that we can use a simple username/password for test/development, and set a secure login for the production environment. On startup the server should check if the user exists in the database, and if they don't, it should create the user.

FGRCL:
# General info

**Issue number**: #272 

**Task description**: 

 - Added endpoint to authenticate in the backend
 - Implemented the frontend to use authentication properly
 
 
| Original   | Updated|
| | |

# Testing
Changed the integration tests to use authorization

**Test coverage**:

# Additional notes

**Note to reviewers**:
To authenticate with the API the user must use HTTP basic authentication on the endpoint "auth/token", they'll receive a token in the "Authorization" header. For other endpoints, the token must be sent in the "Authorization" header with "Bearer".

**To do before merging**:


ButlerJAL (commented):
@FGRCL, quickly noted, can you fix the 13 code smells reported? I've looked at them, and they should be super easy to fix. Ten code smells are in the frontend which you can fix them manually by adding a few whitespaces here and there, or running Prettier should do the trick. Three code smells are in the backend regarding import ordering.

FGRCL:
@MrJCipherButtles OK! smelly code is gone. The google-java-formatter was reverting my eslint fixes in the background every time I was committing.

src/main/resources/application.properties
```diff
@@ -1,6 +1,10 @@
 # Use the development profile to run application using a local MySQL server
-spring.profiles.active=production
```
felixlapierre: Why remove this? Don't we need this for it to run as production in deployment?

src/test/java/features/RunCucumberTest.java
```diff
@@ -1,6 +0,0 @@
-package features;
-
-import io.cucumber.junit.platform.engine.Cucumber;
-
-@Cucumber
-public class RunCucumberTest {}
```
felixlapierre, line 6: Why delete this?

src/main/resources/application.properties
FGRCL: That's probably a mistake

src/test/java/features/RunCucumberTest.java
FGRCL: We already have `CucumberTest.java` in the same folder. I think we introduced that duplicate somewhere along the line. The cucumber tests still run fine though.

src/main/resources/application.properties
FGRCL: ok it's back in

ButlerJAL (approved):
I've reviewed the security hotspots. All looks safe to me. Using the Authorization header with the Bearer authentication scheme to send JWT is top notch. Not using session cookies mitigates CSRF attacks. LGTM!

felixlapierre:
This also closes #278 and #279 btw

PR #316, opened 2021-02-21
[#26] Rename device

### LINKED ISSUE #26: Rename a device
When viewing a device, the user should be able to edit certain information fields. Probably only the name will be editable for now, since device information pretty much depends on the capabilities of the device.

** Acceptance Criteria **
- There is an edit button or pencil button on the device detailed information page
- When clicking that button, the "name" field becomes editable
- After editing and pressing enter or a save button, an update request is sent to the server to rename the device.
- The new name of the device is now shown in the frontend

### LINKED ISSUE #317: Calling PUT /device returns a 500 error if the optional "status" field is not included
**Steps to reproduce:**
- Register a device in the system
- Send a PUT request to /device with a body including the serial number of the device and a new display name

Expected result: Successful request, with a response containing the device's fields

Actual result: 500 error

**Cause**:
- DeviceAspect will check the device status in order to log if the device is going online or offline. If the status was not provided, this throws a null pointer exception. 

**Fix**:
- Since we aren't going to create the device status logs through the "update" command anymore anyways, I changed that aspect to simply log that the device was updated.

gdzooks:
# General info

**Issue number**: #26 

**Task description**: 
 - There is an edit button or pencil button on the device detailed information page
 - When clicking that button, the "name" field becomes editable
 - After editing and pressing enter or a save button, an update request is sent to the server to rename the device.
 - The new name of the device is now shown in the frontend
 
 
| Original   | Updated|
| image | image |
| N/A -> more updated steps | image |
| N/A -> more updated steps | image |
| N/A -> more updated steps | image |

# Testing

**Test coverage**:

- 100%!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

# Additional notes
**Note to reviewers**:

- not a special note but testing is truly hell

**To do before merging**:
- [x] Bug for refresh because of `{location: {state: device}}`
  - bug here: #318


felixlapierre:
Closes #317 
This PR includes a change in the backend that fixes that but which was preventing this feature from working.

felixlapierre:
Also I can validate that this works with the backend and passes the acceptance criteria ✔️ will review the code tmrw

felixlapierre:
There's an issue right now where sending the PUT request to the server without including all the fields of the device object will erase the fields not provided. This causes crashes and all kinds of bad stuff. I'll look into fixing this in the backend (hopefully today), because it was something we wanted to support for performance reasons, but until then let's not merge this PR. The code can be reviewed tho.

frontend/src/deviceDetailsPage/DeviceNameDetail.jsx
```diff
@@ -0,0 +1,117 @@
...
+            </Box>
+            <Box padding={4} paddingLeft={0}>
+              <Button
+                id="confirmEditBtn"
+                type="submit"
+                color="primary"
+                variant="contained"
+                disableElevation
+              >
+                Save Changes
```
sharon-ho: I would suggest just calling it `Save` instead of `Save Changes`

frontend/src/deviceDetailsPage/DeviceNameDetail.jsx
```diff
@@ -0,0 +1,117 @@
...
+  setName(updatedName) {
+    this.setState({ name: updatedName });
+  }
+
+  startEdit() {
+    this.setState({ editing: true });
+  }
+
+  cancelEditing() {
+    this.setState({ name: this.deviceName });
```
sharon-ho, line 32: Why is this setState for name necessary when you aren't changing the name of the device? Is it because of the onChange to the text field target value it temporary sets the name state to the value in the text field but if you cancel you want to reset it to the original device name before editing was initiated?

sharon-ho (dismissed):
Looks good! Just one slight suggestion and question for clarification.

frontend/src/deviceDetailsPage/DeviceNameDetail.jsx
gdzooks: more or less yeah! since we don't have refresh happening on the page but still want to show the updated name after changes have been saved, we use the name saved in the component's state for display 
that meant that if you made a change and then pressed cancel, the device name would end up displaying the changes you meant to discard, so i added this line :)

frontend/src/deviceDetailsPage/DeviceNameDetail.jsx
gdzooks: addressed 3e16b27

sharon-ho (approved):

felixlapierre:
We have a hotfix incoming for the update issue so i'm good to merge this

felixlapierre (approved):

PR #319, opened 2021-02-21
[#220] Implement an endpoint to create logs

### LINKED ISSUE #220: Endpoint for devices to report errors during a stream
Create an endpoint for devices to report when an error happened in the stream.
This should create a normal log object (with generated id, message representing what happened, the device that created the log, time at which the log was generated, log level).

sebhani:
**Issue number**: #220 

**Task description**: 

 - Update the contracts to have a CreateLogRequest class and a post endpoint for creating a log (/logs)
 - Override the controller method createLog within the LogController class
 - Add additional mappers, one of them is a mapper to map a String to OffsetDateTime
 - Add createLog method in the LogDaoImpl class
 - Add two tests for the service and controller
- Add two tests for default method map in LogMapper

# Additional notes

**Note to reviewers**:
- dateTime attribute in the CreateLogRequest accepts the following format: YYYY-MM-DDTHH:mm:ss+00:00. For example, "2021-02-04T08:21:23+00:00" is an accepted value.
- Here's a screenshot for the request to ease your revision process:
postman request for creating a new log

### DIFF
--- a/src/main/java/org/beanpod/switchboard/controller/LogController.java
+++ b/src/main/java/org/beanpod/switchboard/controller/LogController.java
@@ -7,6 +7,7 @@ import lombok.RequiredArgsConstructor;
 import org.beanpod.switchboard.dao.LogDaoImpl;
 import org.beanpod.switchboard.dto.mapper.LogMapper;
 import org.beanpod.switchboard.exceptions.ExceptionType;
+import org.beanpod.switchboard.service.LogService;
 import org.openapitools.api.LogsApi;
 import org.openapitools.model.CreateLogRequest;
 import org.openapitools.model.LogModel;
@@ -21,6 +22,7 @@ public class LogController implements LogsApi {
   public static final String CONTROLLER_NAME = "Log";
   private final LogDaoImpl logDao;
   private final LogMapper logMapper;
+  private final LogService logService;
 
   @Override
   public ResponseEntity<List<LogModel>> retrieveAllLogs() {
@@ -35,10 +37,9 @@ public class LogController implements LogsApi {
   @Override
   public ResponseEntity<LogModel> createLog(@Valid CreateLogRequest createLogRequest) {
     return Optional.of(createLogRequest)
-        .map(logMapper::toLogDto)
-        .map(logDao::createLog)
-        .map(logMapper::toLogEntity)
-        .map(logMapper::toLogModel)
+        .map(logMapper::createLogRequestToLogModel)
+        .map(logService::createLog)
+        .map(logMapper::logDtoToLogModel)
         .map(ResponseEntity::ok)
         .orElseThrow(() -> new ExceptionType.UnknownException(CONTROLLER_NAME));
   }
--- a/src/main/java/org/beanpod/switchboard/controller/LogController.java
+++ b/src/main/java/org/beanpod/switchboard/controller/LogController.java
@@ -4,9 +4,7 @@ import java.util.List;
 import java.util.Optional;
 import javax.validation.Valid;
 import lombok.RequiredArgsConstructor;
-import lombok.extern.java.Log;
 import org.beanpod.switchboard.dao.LogDaoImpl;
-import org.beanpod.switchboard.dto.LogDto;
 import org.beanpod.switchboard.dto.mapper.LogMapper;
 import org.beanpod.switchboard.exceptions.ExceptionType;
 import org.openapitools.api.LogsApi;
@@ -19,6 +17,7 @@ import org.springframework.web.bind.annotation.RestController;
 @RestController
 @RequiredArgsConstructor
 public class LogController implements LogsApi {
+
   public static final String CONTROLLER_NAME = "Log";
   private final LogDaoImpl logDao;
   private final LogMapper logMapper;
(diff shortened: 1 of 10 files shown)


mohhef (changes requested):
@sebhani @felixlapierre is it possible to get some backstory on how we would need/use this?, lets say there is a crash in the stream from the BE side, would the FE then be calling this endpoint? I am a little confused

src/main/java/org/beanpod/switchboard/dto/mapper/LogMapper.java
```diff
@@ -11,4 +14,17 @@
   List<LogModel> toLogModels(List<LogEntity> logEntityList);
 
   LogModel toLogModel(LogEntity logEntity);
+
+  LogEntity toLogEntity(LogDto logDto);
+
+  LogDto toLogDto(CreateLogRequest createLogRequest);
+
+  LogDto toLogDto1(LogEntity logEntity);
```
FGRCL: let's rename this to something like `logEntityToDto

src/main/java/org/beanpod/switchboard/dto/mapper/LogMapper.java
```diff
@@ -11,4 +14,17 @@
...
+
+  /* to convert string to OffsetDateTime
+   * value must be of YYYY-MM-DDTHH:mm:ss+00:00
+   */
+  default OffsetDateTime map(String value) {
+    return OffsetDateTime.parse(value);
+  }
```
FGRCL, line 33: This would fit well in `util/DateUtil.java`

src/main/java/org/beanpod/switchboard/controller/LogController.java
```diff
@@ -24,4 +31,15 @@
   public ResponseEntity<List<LogModel>> retrieveDeviceLogs(@PathVariable String serialNumber) {
     return ResponseEntity.ok(logDao.getDeviceLogs(serialNumber));
   }
+
+  @Override
+  public ResponseEntity<LogModel> createLog(@Valid CreateLogRequest createLogRequest) {
+    return Optional.of(createLogRequest)
+        .map(logMapper::toLogDto)
+        .map(logDao::createLog)
```
FGRCL: Ideally, the request would go through `LogService` first.

src/test/java/org/beanpod/switchboard/mapper/LogMapperTest.java
```diff
@@ -0,0 +1,47 @@
...
+    try {
+      OffsetDateTime offsetDateTime = logMapper.map(validDateTime);
+      Assertions.fail("Expected exception to be thrown");
+    } catch (DateTimeParseException e) {
+      assertThat(e)
+          .isInstanceOf(DateTimeParseException.class)
+          .hasMessage("Text '2017-05-12 05:45+00:00' could not be parsed at index 10");
+    }
+  }
+}
```
FGRCL, line 47: If you decide to move `LogMapper.map` theses tests would go in `DateUtilTest`

FGRCL (commented):
I've got small changes to suggest. Apart from that, it's pretty textbook 👍

src/main/java/org/beanpod/switchboard/dto/mapper/LogMapper.java
sebhani: I don't think it's appropriate to be in DateUtil class since the class needs to be mapper (@Mapper). I found two people so far did the same way I did: https://stackoverflow.com/questions/59616631/mapping-2-string-fields-to-offsetdatetime-with-mapstruct

An alternative would be extracting this method to a new class, but I don't see why would we do that.

src/main/java/org/beanpod/switchboard/dto/mapper/LogMapper.java
sebhani: Thank you for the suggestion, I have updated the method signature.

src/main/java/org/beanpod/switchboard/controller/LogController.java
```diff
@@ -24,4 +31,15 @@
...
+        .map(logMapper::toLogEntity)
+        .map(logMapper::toLogModel)
```
mohhef: why do we log to entity and then to model cant we directly toLogModel

mohhef (commented):
LGTM, i have 1 question

src/main/java/org/beanpod/switchboard/controller/LogController.java
sebhani: We want to go from LogDto to LogModel, but toLogModel goes from LogEntity to LogModel

mohhef: why no create a new method?

LogModel toLogModel(LogDto x)

sebhani: Good suggestion to remove middle mappings. I have updated the code accordingly.

src/main/java/org/beanpod/switchboard/controller/LogController.java
sebhani: I have updated the code according to your suggestion.

mohhef (approved):
LGTM. Comments addressed

src/main/java/org/beanpod/switchboard/dto/mapper/LogMapper.java
FGRCL: I see, I missed that the first time I read.

FGRCL (approved):

felixlapierre:
Question: does this let devices create logs that are related to a particular stream? I.E. create an error log related to stream X with a specific message indicating what the error was. That way when we got logs for that stream we will see the error. Or will that be in a later PR?

sebhani:
@felixlapierre No, this doesn't create a  log related to a specific stream since there's no relation between log_entity and stream. It should be an easy change to implement imo.

PR #323, opened 2021-02-21
Performance optimization: Allow partial updates

### LINKED ISSUE #324: Peformance testing optimization
From our performance testing analytics we noticed that update requests take a lot of time.
We would like to allow the user to update partial fields and not put the whole request body while only updating one field

**Acceptance criteria**
Allow users to update Devices without the need to put all the body parameters in the request body

mohhef:
# General info
This pr optimizes requests by allowing partial updates in the database.


felixlapierre (approved):

felixlapierre (approved):
Tested on my end and it works great!

ButlerJAL (approved):

PR #331, opened 2021-02-23
Refactor streamlist

jun-loke:
# General info

**Issue number**: N/A

**Task description**: 

- Refactor StreamList and friends so that there is a wrapper function that handles the onChange events
- Remove tests on StreamList that weren't testing Streamlist
- Rewrite tests for StreamList
- Write test for StreamsTableWrapper
- Write tests for StreamsTable
- 
**Update**: because the StreamList page's test file was testing the inner components and not the Streamlist, fhe file was rewritten to test only the Streamlist. However, this decreased the total code coverage from 95.1% to ~94.4% because this is a regression, additional tests and refactors will be added to this pr. However it's already proven to be a lot of work and a lot of new code. I kinda want to put it in its own pr
- Refactor delete stream Dialog & add tests
 
 
| Original   | Updated|
|:---------------:|:---------------:|
| | |

# Testing

**Test coverage**:

# Additional notes

**Note to reviewers**:

The Streamlist test was testing the StreamsTable component within it and not actually the StreamList. Those tests have been removed. 
**To do before merging**:
- [ ] Edit Changelog(if necessary)

PR #332, opened 2021-02-23
Refactor Dialog

### LINKED ISSUE #333: Delete stream test coverage lacking
While doing some cleanup in #331 I found that the testing for the `DeleteStream `Dialog was not in a corresponding test file, but was being tested in `StreamList.test.js` I cleaned up that file in that PR, however this lowers our total cover coverage by 1%. 

This 1% needs to be recovered.

jun-loke:
# General info

**Issue number**: Closes #333 
Making this PR to check what lines of code the `Streamlist.test.js` file is covering due to refactor in pr #331 
From what I can read, I see that it's covering the dialog, but a drop of 1% is significant Also refactoring the Dialog
**Task description**: 

 - Move Dialog to composed components
 
| Original   | Updated|
| | |

# Testing

**Test coverage**:

# Additional notes

**Note to reviewers**:

**To do before merging**:


frontend/src/general/dialog/DeleteStreamDialogSummoner.jsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { IconButton, Tooltip } from "@material-ui/core";
+import { Delete } from "@material-ui/icons";
+import { withRouter } from "react-router-dom";
+
+import DeleteStreamDialog from "./DeleteStreamDialog";
+
+class DeleteStreamDialogSummoner extends React.Component {
```
felixlapierre: I would suggest `DeleteStreamDialogOpener` instead of Summoner. Since summoning isn't a technical term it's unclear from the name whether or not this creates the dialog or just opens it.

frontend/src/general/dialog/DeleteStreamDialog.jsx
```diff
@@ -0,0 +1,61 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import Dialog from "./Dialog";
+import { deleteStream } from "../../api/StreamApi";
+
+export default class DeleteStreamDialog extends React.Component {
```
felixlapierre: IMO this component belongs in a folder with all the stream-related components, not in the general folder.

frontend/src/general/dialog/DialogBody.jsx
```diff
@@ -0,0 +1,22 @@
...
+export default function DialogBody(props) {
+  const { children } = props;
+  return (
+    <MuiDialogContent>
+      <MuiDialogContentText id="dialog-description">
+        {children}
+      </MuiDialogContentText>
+    </MuiDialogContent>
+  );
+}
```
felixlapierre, line 15: Do we really need a component for this? Can we just put the `<MuiDialogContent>` and `<MuiDialogContentText>` directly in the `Dialog` element since there is no other logic here?

frontend/src/general/dialog/DialogTitle.jsx
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import PropTypes from "prop-types";
+import MuiDialogTitle from "@material-ui/core/DialogTitle";
+
+export default function DialogTitle(props) {
+  const { children } = props;
+  return (
+    <MuiDialogTitle id="delete-stream-dialog-title">{children}</MuiDialogTitle>
+  );
+}
```
felixlapierre: I feel like this component could also be inlined in the dialog component since it contains no logic, only a MuiDialogTitle component. Also, the ID should not be `delete-stream-dialog-title` for a generic component.

frontend/src/general/dialog/__tests__/DeleteStreamDialog.test.jsx
```diff
@@ -0,0 +1,78 @@
...
+    });
+  });
+  describe("confirmDelete() function", () => {
+    it("calls StreamAPI with the passed ID", () => {
+      wrapper.instance().confirmDelete();
+      expect(StreamApi.deleteStream).toBeCalledWith(dummyId);
+    });
+  });
+  describe("afterDelete() function", () => {
+    it("calls the correct functions", () => {
```
felixlapierre: Rename to describe better what it calls (e.g. Closes the dialog and refreshes the page)

felixlapierre (commented):
A couple of suggestions otherwise LGTM

frontend/src/general/dialog/DialogBody.jsx
jun-loke: I prefer having it in its own component because it allows our dialog element to be cleaner, requiring only one tag pair instead of two. This also prevents Dialog from growing extremely large

frontend/src/general/dialog/__tests__/DeleteStreamDialog.test.jsx
jun-loke: fixed in 7af4db61

frontend/src/general/dialog/DeleteStreamDialog.jsx
jun-loke: 4851bbb1

frontend/src/general/dialog/DeleteStreamDialogSummoner.jsx
jun-loke: 44b9153f

frontend/src/general/dialog/DialogTitle.jsx
jun-loke: k

jun-loke: 54c634e

gdzooks (approved):
this was what i wanted to do at the start but had too much trouble getting the states and hook stuff to work right. big fan! very excited that jun has it up and running! 👍

one comment i'd like a reply on (button1 & button2 names) but otherwise very delectable PR :yum:

frontend/src/general/dialog/Dialog.jsx
```diff
@@ -0,0 +1,61 @@
...
+      <MuiDialog
+        open={open}
+        onClose={() => this.closeDialog}
+        aria-labelledby="dialog-title"
+        aria-describedby="dialog-description"
+        id="dialog"
+      >
+        <DialogTitle>{title}</DialogTitle>
+        <DialogBody>{children}</DialogBody>
+        <DialogButtons button1={cancelButton} button2={actionButton} />
```
gdzooks, line 45: ~~my comments didnt post earlier 😫~~
are `button1` and `button2` the best names here? if our designs are consistently putting the cancel button on the left as button1 then it might be clearer to change the names to `cancelButton` and `actionButton` or something

frontend/src/general/dialog/DialogBody.jsx
gdzooks: it's nice to have the test files smaller as well ^

frontend/src/streamlist/DeleteStreamDialogOpener.jsx
```diff
@@ -0,0 +1,48 @@
...
+          ref={this.dialogElement}
+          deleteId={deleteId}
+          history={history}
+        />
+      </>
+    );
+  }
+}
+
+export default withRouter(DeleteStreamDialogOpener);
```
gdzooks, line 40: looked up wht this does and NICE 👌

frontend/src/general/dialog/Dialog.jsx
jun-loke: Since the dialogButtons are technically their own component, the idea was to leave that flexibility there, though it's unlikely that we will need to reuse it. 
But I do agree the name is a bit weak. Maybe we can rename it to leftButton and rightButton?

felixlapierre (approved):
Comments addressed LGTM

PR #339, opened 2021-02-25
[#267] Endpoint for statistics

### LINKED ISSUE #267: Endpoint for receivers to send SRT stream statistics to the server
Once #269 is completed and we know what statistics we want to track for SRT, we should create an endpoint for devices to provide statistics about a stream. 

Possible designs:

* Have the statistics as a property of the stream object, and simply call PUT /streams/{id}
* Have the statistics in their own table and POST/PUT a statistics object, and have a one-to-one relation between streams and statistics
* Have the statistics as a property of the stream, but they must be created/updated using a POST/PUT /streams/{id}/statistics

Also, what do we do when there are no statistics provided yet? Devices shouldn't have to do POST the first time they push statistics and PUT after that; they should just send their statistics and let the server record them.

We should also create a GET endpoint for these statistics to be retrieved in the frontend to be displayed.

mohhef:
# General info
Workflow:
- Create a stream between two channels
- A stream stat object for that stream will be automatically created
- Now you can update that stream stat object
- You can get all the stream stats

**Issue number**: 257

- [x]  Unit tests


src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
```diff
@@ -52,6 +70,17 @@ public StreamEntity updateStream(StreamDto streamDto) {
     return streamRepository.save(streamEntity);
   }
 
+  public StreamStatDto updateStreamStat(StreamStatDto streamStatDto) {
+    if (!streamRepository.existsById(streamStatDto.getId())) {
+      throw new StreamDoesNotExistException(streamStatDto.getId());
+    }
+    Optional<StreamStatDto> streamStat = getStreamStat(streamStatDto.getId());
+    statMapper.updateStreamStatFromDto(streamStatDto, streamStat.orElse(null));
```
felixlapierre, line 78: Can you make it accept partial updates like we did for update device? We might end up getting stats from the sender and the receiver, so each will be pushing only some of the values.

contracts/api-spec.yaml
```diff
@@ -613,6 +650,82 @@ components:
           format: int64
           example: '2'
 
+    StreamStatModel:
+      description: A stat that is related to a specific stream
+      properties:
+        id:
+          type: integer
+          format: int64
```
felixlapierre, line 710: Some of these properties can have decimal values, so integer type won't work for them

* link rtt
* link bandwidth
* send mbitRate
* recv mbitRate

felixlapierre, line 710: Should we instead have nested properties like the json output of srt-live-transmit? If that's too much of a pain to store in the database, can we reformat at the service layer or something?
image

contracts/api-spec.yaml
```diff
@@ -407,6 +407,37 @@ paths:
...
+      description: Retrieve all the existing stream statistics
+      tags:
+        - Streams
+      responses:
+        '200':
+          description: Successful retrieval of stream statistics
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/StreamStatsModel'
```
felixlapierre, line 454: I realized I misunderstood a question you asked me about this - we need an endpoint where we retrieve the stats of a single stream. We aren't really ever going to need to get all of the stream statistics (although we can leave it in if we want)
Something like GET /stream/1/statistics or /stream/statistics/1

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
```diff
@@ -37,7 +43,19 @@ public StreamDto saveStream(StreamDto streamDto) {
...
+    StreamDto streamDto1 = mapper.toDto(streamRepository.save(streamEntity));
+
+    // Save an empty stream stat when saving a stream
+    if (streamDto.getStreamStat() == null) {
+      StreamStatEntity streamStatBuild =
+          StreamStatEntity.builder().stream(mapper.toEntity(streamDto1))
+              .id(streamDto1.getId())
+              .build();
+      streamStatRepository.save(streamStatBuild);
+      streamDto1.setStreamStat(statMapper.toDto(streamStatBuild));
```
felixlapierre, line 55: What does this do since the streamDto has already been saved without the stream stats?

felixlapierre (changes requested):

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
mohhef: Yes it does  accept partial updates

contracts/api-spec.yaml
mohhef: fixed

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
mohhef: it checks when you create a stream object, are you saving a stream stat object along side with it on the create request? No? then create an empty streamstat object and save it to the DB.
Yes? then good the stream + the stat should get saved

contracts/api-spec.yaml
mohhef: given that i save all the attributes in one table to not complicate things, i think it would add a bit of boiler plate code if i would to do them as seprate objects

mohhef: let me know if you  have idea

contracts/api-spec.yaml
mohhef: fixed

contracts/api-spec.yaml
mohhef: i think maybe i can just create nested objects or something, ill try that tomorrow

src/main/java/org/beanpod/switchboard/dao/StreamDaoImpl.java
mohhef: You were right it didnt acccept partial values although i thought it was,
thats because i used long instead of Long. Now it works!

contracts/api-spec.yaml
mohhef: Your idea was good, i applied it. Now it should be good

felixlapierre: Works great now! 😄

felixlapierre (approved):
My suggestions were addressed, LGTM!

gdzooks (approved):
looks good ^q^

PR #349, opened 2021-03-01
[#334] StreamApi now uses promises. Improved StreamApi tests.

### LINKED ISSUE #334: Stream API missing tests
As discovered in #331 the `StreamAPI` was being tested in `StreamList.test.js` and has no actual individual testing.

felixlapierre:
# General info

**Issue number**: #344 

**Task description**: 

 - Improve StreamApi tests to actually test the functionality.
 - Refactor StreamApi to use a promise-based async mechanism rather than a callback-based async mechanism.
 
 

# Testing

Although test coverage might not increase, this will increase how much behaviour is actually validated.


frontend/src/streamlist/DeleteStream.jsx
```diff
@@ -24,7 +24,7 @@ export default function DeleteStream(props) {
     return setOpen(false);
   };
   const confirmDelete = () => {
-    StreamApi.deleteStream(deleteId, () => {
+    StreamApi.deleteStream(deleteId).then(() => {
       history.push("/Streaming");
       history.go(0);
```
jun-loke: Delete stream file is deleted in #332 You can also delete it here if that helps with your test coverage numbers

frontend/src/api/tests/StreamApi.test.js
```diff
@@ -2,56 +2,163 @@ import axios from "axios";
...
-const respAllStreams = {
-  target: {
-    streams: mockStreams
-  }
-};
-const respSingleStream = {
-  target: {
-    stream: mockSingleStream
-  }
+const axiosGetSecondStreamResponse = {
```
jun-loke: Do you actually need all this information? Since it's a mock call, can't you just make an object with the bare minimum properties?
getStream only uses stream id and the encoder & decoder. Thus you can simply have 
const dummyStream={
name:"test",
outputChannel: { encoder: {...}},
inputChannel: {decoder:{...}}
}

And you're mocking the return value anyways. You aren't using most of these fields

jun-loke (changes requested):
looks fine. Just a small thing with dummy data

frontend/src/api/StreamApi.js
```diff
@@ -1,45 +1,31 @@
...
-    .catch(() => {
-      SampleData.getAllStreams(callback);
-    });
+export async function getAllStreams() {
+  return axios.get(process.env.REACT_APP_STREAM).then((streams) => {
+    return Promise.all(
+      streams.data.map((streamId) => {
+        return getStream(streamId);
+      })
+    );
```
FGRCL, line 25: I think you could get away with creating an array of Streams then returning that instead of the Promise.all.

frontend/src/api/tests/StreamApi.test.js
```diff
@@ -2,56 +2,163 @@ import axios from "axios";
...
+        serialNumber: "d7TxFn7o",
+        publicIpAddress: "0:0:0:0:0:0:0:1",
+        privateIpAddress: "127.0.0.1",
+        displayName: "sample_sender",
+        status: "online"
+      },
+      output: null
+    }
+  },
+  isRendezvous: false
```
FGRCL: Maybe we should  start writing fixtures for the frontend tests. The Sample data could be repurposed for that.

frontend/src/api/tests/StreamApi.test.js
```diff
@@ -2,56 +2,163 @@ import axios from "axios";
...
+    it("should call axios.get and return a single stream's information", async () => {
+      axios.get.mockResolvedValue({ data: axiosGetFirstStreamResponse });
+      const response = await StreamApi.getStream(123);
 
       // check that callback was invoked with correct value
       expect(axios.get).toHaveBeenCalledWith(
         `http://localhost:8080/stream/123`
       );
+
+      expect(response).toEqual(expectedFirstStreamRespone);
```
FGRCL, line 49: How come the response is equal to an object here and not a Promise?

frontend/src/api/tests/StreamApi.test.js
felixlapierre: I included all the fields returned by the backend in order to follow the principle that test data should reflect production data as closely as possible. Also, I suspect we'll use the full stream information when we create the detailed stream information page.

frontend/src/streamlist/DeleteStream.jsx
felixlapierre: Rebased with master to get the changes from the refactor dialog PR, and the coverage has indeed increased 😄

frontend/src/api/StreamApi.js
felixlapierre: Since getStream returns a promise (as it calls the backend) I don't really see a way to write this code without using `Promise.all`. This implementation waits for all the backend calls to complete and then creates an array from the returned streams. Could you maybe show me the code of the alternate implementation you suggest?

frontend/src/api/tests/StreamApi.test.js
felixlapierre: We already have a DeviceFixture.js, I'll create a StreamFixture.js to house this.

frontend/src/api/tests/StreamApi.test.js
felixlapierre: Because i `await`ed the return value of `getStream`.

frontend/src/api/tests/StreamApi.test.js
```diff
@@ -2,56 +2,81 @@ import axios from "axios";
...
+const { secondStreamResponse } = StreamFixture;
+const { allStreamsResponse } = StreamFixture;
+
+const expectedFirstStreamRespone = new StreamInfo(
+  1,
+  convertToDataObject(firstStreamResponse.outputChannel.encoder),
+  convertToDataObject(firstStreamResponse.inputChannel.decoder),
+  ["Additional stream info goes here."]
+);
+const expectedSecondStreamRespone = new StreamInfo(
```
jun-loke: typo

felixlapierre: Fixed in 6f30b79!

frontend/src/api/StreamApi.js
FGRCL: On second thought I don't think it'll work, I didn't realize getStream returned a promise.

frontend/src/api/tests/StreamApi.test.js
FGRCL: makes sense

frontend/src/api/tests/StreamFixture.js
```diff
@@ -0,0 +1,87 @@
...
+        publicIpAddress: "0:0:0:0:0:0:0:1",
+        privateIpAddress: "127.0.0.1",
+        displayName: "sample_sender",
+        status: "online"
+      },
+      output: null
+    }
+  },
+  isRendezvous: false
+};
```
FGRCL, line 87: I like!

FGRCL (dismissed):
LGTM

frontend/src/api/tests/StreamApi.test.js
```diff
@@ -2,56 +2,81 @@ import axios from "axios";
...
-  }
-};
-const respSingleStream = {
-  target: {
-    stream: mockSingleStream
-  }
-};
+const { firstStreamResponse } = StreamFixture;
+const { secondStreamResponse } = StreamFixture;
+const { allStreamsResponse } = StreamFixture;
```
jun-loke: nitpick: this could be one line
```suggestion
const { firstStreamResponse, secondStreamResponse, allStreamsResponse } = StreamFixture;
```

jun-loke (dismissed):

jun-loke (approved):
looks good

FGRCL (approved):
Reapproving

PR #353, opened 2021-03-03
[#250] Delete device bug

### LINKED ISSUE #250: Internal Server Error when deleting a device that has active streams
Steps to reproduce:
- Start the frontend, backend, sample sender, and sample receiver
- Click the register button on both sample apps
- Open the admin panel
- Open the streams page
- Create a stream between both sample apps
- Open the device list page
- Click on one of the devices to open the detailed device information page
- Click on the delete button

Expected result:
- A prompt appears "This device is involved in active streams; if you delete this device, all streams involving this device will be deleted. Are you sure you want to continue?"
- Clicking confirm redirects to the device list page
- The device should no longer be present in the device list page

Actual result:
- The confirmation prompt does not mention any active streams
- The request to the server is met with a 500 error

- The user stays on the device details page, and the device is still present in the system

mohhef:
# General info
When deleting a device, it deletes everything associated with it.


felixlapierre (approved):
Validated that this works. LGTM!!

sebhani (approved):
Works fine!

PR #354, opened 2021-03-03
[#325] Add roles in user management

### LINKED ISSUE #325: Add roles in user management
Only difference between regular users and admin users is that admin users can create new accounts and view accounts in the system.

ButlerJAL:
# General info

**Issue number**: #325 

**Task description**: 

 - Configure Spring Security and JWT to play nice together regarding user Roles and Permissions for resource access
 - ~Started investigating updating our system tests to authenticate (To be continued)~ (out of PR scope)
 - Renamed /logs endpoint to /log to be consistent with singular naming
 - Updated frontend LogApi for authentication

# Testing

**Test coverage**: n/a

# Additional notes

**Note to reviewers**:

~/Log endpoint is returning a 401 Unauthorized response instead of a list of logs ???~ (fixed)

**To do before merging**:


ButlerJAL:
@FGRCL /Log endpoint is returning a 401 Unauthorized response instead of a list of logs. Can you see why? I've been looking at the code, and I don't see why this behaviour is happening. The /Log endpoint should fall within the scope of `.antMatchers("/**").hasAnyAuthority("USER", "SUPERUSER")`which should behave like the other endpoints when requested by a valid USER.

pom.xml
```diff
@@ -155,6 +155,20 @@
...
       <version>${java-jwt-version}</version>
     </dependency>
+    <!--
+    <dependency>
+      <groupId>org.testng</groupId>
+      <artifactId>testng</artifactId>
+      <version>7.4.0</version>
+      <scope>test</scope>
+    </dependency>
+    -->
```
FGRCL: If we don't need this depedency anymore could we remove it?

ButlerJAL:
> @FGRCL /Log endpoint is returning a 401 Unauthorized response instead of a list of logs. Can you see why? I've been looking at the code, and I don't see why this behaviour is happening. The /Log endpoint should fall within the scope of `.antMatchers("/**").hasAnyAuthority("USER", "SUPERUSER")`which should behave like the other endpoints when requested by a valid USER.

FIXED

src/main/java/org/beanpod/switchboard/config/WebSecurityConfig.java
```diff
@@ -39,11 +41,16 @@ protected void configure(HttpSecurity http) throws Exception {
         .sessionCreationPolicy(SessionCreationPolicy.STATELESS)
         .and()
         .authorizeRequests()
-        .antMatchers("/user/sign-up")
-        .permitAll()
+        .antMatchers("/user/**")
+        .hasAnyAuthority("ADMIN", "SUPERUSER")
+        .antMatchers("/**")
+        .hasAnyAuthority("USER", "SUPERUSER")
```
jun-loke, line 47: What's the difference between the `hasAnyAuthority("ADMIN", "SUPERUSER")`  and `.hasAnyAuthority("USER", "SUPERUSER")` calls here?  rather, why is SuperUser in both?

jun-loke (dismissed):
Just one point of clarification, but from what I can see looks good

src/main/java/org/beanpod/switchboard/config/WebSecurityConfig.java
ButlerJAL: `SUPERUSER`: is a user with access to everything. It's to be used solely for development purposes.
`ADMIN`: is a user with access to only the user controller which allows them to manage/signup users.
`USER`: is a user with access to everything except the user controller.

jun-loke: wait, isn't `SuperUser `actually going into the app? The PM clarified a bit ago that our Admin user would have access to the whole app + user creation

ButlerJAL: Yes, `SUPERUSER` reflects that requirement in which we have an app user with the user creation privilege; whereas, `ADMIN` is a non-user with only user creation privilege

ButlerJAL: Should I replace `SUPERUSER` with `ADMIN`, and remove the `SUPERUSER` role? Adding/removing roles is fairly straightforward. This can be done quickly before we merge.

src/main/java/org/beanpod/switchboard/config/CustomAuthenticationEntryPoint.java
```diff
@@ -0,0 +1,37 @@
...
+import java.io.PrintWriter;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.web.AuthenticationEntryPoint;
+import org.springframework.stereotype.Component;
+
+@Component
+public class CustomAuthenticationEntryPoint implements AuthenticationEntryPoint {
```
FGRCL, line 15: Not sure what's the role of this filter is,  it seems to reject all requests it receives?

FGRCL: After reading below I understand it handles bad authentication. what do you think about renaming this something like `BadAuthorizationFilter` or `InvalidLoginHandler`

src/main/java/org/beanpod/switchboard/service/UserService.java
```diff
@@ -15,7 +14,6 @@
 public class UserService implements UserDetailsService {
 
   private final UserDaoImpl userDao;
-  private final UserRepository userRepository;
```
FGRCL, line 18: Strange, was that unused on master?

src/main/java/org/beanpod/switchboard/config/WebSecurityConfig.java
FGRCL: I like having the granularity, we just have to make sure that everyone knows the difference since there we sort of changed the naming between the requirements and the implementation.

felixlapierre: I think we can keep it like this. We have the roles we need (user and superuser). I think the admin role makes sense to have even though the PM won't use it right away, they might need it in the future.

felixlapierre (approved):
Validated on my end and the code works great. LGTM!

FGRCL (commented):
Worked as expected on my end, also thanks for keeping the Logs API up to date. I've got 1 suggestion for you before approving.

src/main/java/org/beanpod/switchboard/service/UserService.java
ButlerJAL: Yes it was unused, but it makes sense. Data access should go through the DAO rather than the repository.

pom.xml
ButlerJAL: This is old code, and no longer applicable. Likely you may have started your review on older code.

src/main/java/org/beanpod/switchboard/config/CustomAuthenticationEntryPoint.java
ButlerJAL: What this does is leverage the spring security framework to send a JSON response when a user is unauthorized. For example, if a client with the role USER tries to do admin functions such as create another user, then instead of getting the generic 403 FORBIDDEN response, I customized it to return a 401 UNAUTHORIZED response.

403 FORBIDDEN is still used for non-users who try to access the service.

ButlerJAL: I don't know if you'll agree, but `CustomAuthenticationEntryPoint` is more appropriate than your suggestion?

pom.xml
FGRCL: You're probably right, I don't remember commenting that

src/main/java/org/beanpod/switchboard/config/CustomAuthenticationEntryPoint.java
FGRCL: I see, let's go with what we have now.

FGRCL (approved):

PR #355, opened 2021-03-03
[#351] Create logs related to streams

### LINKED ISSUE #351: Create logs associated with a stream

sebhani:
**Issue number**: #351 

**Task description**: 

-Refactored the logic for logging. Instead of having a single entity related to all different kinds of logs, I created another entity called StreamLog which has an ISA relation with LogEntity. Any logs related to streams will be stored in Stream Logs having the parent as LogEntity. This way, we don't need to store serial numbers of devices in comma separated manner since this's a wrong approach in relational DBs.

Here's a zoomed-in ER diagram for the new relation:
ER diagram for the relation between StreamLog and LogEntity

# Testing
-To test the functionality, logs must be created when streams are created or deleted (Look at StreamAspect.java)


mohhef (changes requested):
I cant wrap my head around why it was done like that as opposed of what we had, i understand in general comma separated could be bad but we only have a maximum of 2 comma separated attributes that are being persisted to the database.

In your implementation, every time you want to add a stream log, you will create a normal log then you will create a stream log. Thus it defeats the purpose of separating the entities I think. Youd usually want to separate entities to try to avoid repetition
Example:
A student has many books, now if you have books comma separated for a student you will repeat the same books for a lot of students but if you separate them and create a join table in the middle you will avoid repetition 

Also in our case all the data that gets saved to the log entity is through the Aspect which prevents us users from editing the data (i.e. 1,2,helloworld) thus again i think it defeats the purpose of having 2 separate entities.

Not sure if what i said makes sense to you, let me know what you think

felixlapierre (dismissed):

felixlapierre:
Approving since Hani has explained the rationale behind the database structure.

mohhef:
Creating streams results in an unknown error and doesnt enter data into the stream log entity

src/main/java/org/beanpod/switchboard/service/StreamLogService.java
```diff
@@ -0,0 +1,35 @@
...
+  public StreamLogDto createLog(
+      OffsetDateTime dateTime,
+      String message,
+      String decoderSerial,
+      String encoderSerial,
+      String streamId) {
+    LogEntity logEntity = new LogEntity(dateTime, message, "info", decoderSerial);
+    StreamLog streamLog = new StreamLog(encoderSerial, streamId);
+
+    // logEntity.setStreamLog(streamLog); // will cause infinite loop
```
felixlapierre: Remove commented code

src/main/java/org/beanpod/switchboard/dto/LogDto.java
```diff
@@ -23,4 +26,16 @@
...
+  private StreamLog streamLog;
+
+  // for backward compatibility
+  public LogDto(
+      Long id, OffsetDateTime dateTime, String message, String level, String serialNumber) {
+    this.id = id;
+    this.dateTime = dateTime;
+    this.message = message;
+    this.level = level;
+    this.serialNumber = serialNumber;
```
felixlapierre: A way to pull this off without making a custom constructor is to mark all the properties except StreamLog as required, and add the `@RequiredArgsConstructor` annotation. Then it will generate 1 constructor with everything, and 1 constructor with only the required stuff (which is the one you are manually defining here).

src/test/java/org/beanpod/switchboard/dto/LogDtoTest.java
```diff
@@ -0,0 +1,32 @@
...
+    LogDto actualLogDto =
+        new LogDto(
+            LogFixture.id,
+            LogFixture.dateTime,
+            LogFixture.message,
+            LogFixture.level,
+            DeviceFixture.SERIAL_NUMBER);
+
+    assertEquals(logDto, actualLogDto);
+    assertNull(actualLogDto.getStreamLog());
```
felixlapierre: Testing the constructor like this is probably overkill (But i guess we need it for coverage?) Anyways if you replace by `@RequiredArgsConstructor` then you can remove the test

felixlapierre (commented):
Very minor suggestions after that I can give an approve :)

src/main/java/org/beanpod/switchboard/dto/LogDto.java
```diff
@@ -23,4 +26,16 @@
...
+  }
```
mohhef: whats this for

src/main/java/org/beanpod/switchboard/entity/LogEntity.java
```diff
@@ -29,6 +32,17 @@
...
+  @OneToOne(mappedBy = "logEntity", cascade = CascadeType.ALL)
+  @PrimaryKeyJoinColumn
+  private StreamLog streamLog;
+
+  public LogEntity(OffsetDateTime dateTime, String message, String level, String serialNumber) {
+    this.dateTime = dateTime;
+    this.message = message;
+    this.level = level;
+    this.serialNumber = serialNumber;
+  }
```
mohhef: whats this for?

src/main/java/org/beanpod/switchboard/entity/StreamLog.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @OneToOne
+  @MapsId
+  @JoinColumn(name = "log_id")
+  private LogEntity logEntity;
+
+  public StreamLog(String serialNumber, String streamId) {
+    this.serialNumber = serialNumber;
+    this.streamId = streamId;
+  }
```
mohhef: whats this for?

mohhef (changes requested):

src/main/java/org/beanpod/switchboard/entity/LogEntity.java
mohhef: instead of using a constructor, you can use the lombok builder i think

mohhef:
I tested the functionality again.
I either get an 1. internal server error
2. Not an internal server error by the logs dont go into the stream entity

src/main/java/org/beanpod/switchboard/dto/LogDto.java
sebhani: Thanks for letting me know about @RequiredArgsConstructor. I changed my code accordingly.

src/main/java/org/beanpod/switchboard/dto/LogDto.java
sebhani: I believe it's used for creating logs that are not stream-related. If this constructor is commented out, a funky behavior will appear whenever logs are created.

Anyway, i've removed it right now and used lombok annotations to generate it.

src/main/java/org/beanpod/switchboard/entity/LogEntity.java
sebhani: I removed it and used @RequiredArgsConstructor

src/main/java/org/beanpod/switchboard/entity/StreamLog.java
sebhani: it's used for creating StreamLog objects neatly. Anyway, I removed it and used lombok annotations for creating constructors.

mohhef (approved):
looks good to me,
i think if you use the builder instead of constructor youll have less code

felixlapierre (approved):

PR #367, opened 2021-03-06
[#250] Create device bug

mohhef:
# General info
Creating a device without a channel now returns an error


src/main/java/org/beanpod/switchboard/controller/DecoderController.java
```diff
@@ -72,6 +72,9 @@
 
   @PostMapping
   public ResponseEntity<DecoderDto> createDecoder(@RequestBody @Valid DecoderDto decoderDto) {
+    if (decoderDto.getInput().isEmpty()) {
```
sebhani, line 75: Why not trying to use @NotEmpty annotation in the DecoderDto class?

mohhef: doesnt work

mohhef: All javax validations do not work on the set for some reason: notempty & size=1

felixlapierre (approved):

ButlerJAL (approved):

PR #372, opened 2021-03-07
[#330] Add Protected Routes to Front-end

jun-loke:
# General info

**Issue number**: Closes #330  

**Task description**: 
"Redirect to login when not logged in"
### Actual work done

 - Extract <AppRouter/> component from the index.js page for ease of testing
 - Create AdminProtectedRoute and ProtectedRoute components
   - these redirect when not authorized
 - Add cookie handling authentication to `AuthenticationAPI`
   - cookies allow a session to persist within  browser even when the page is refreshed
   - cookies only persist for an hour (easily changeable)
   - has admin route handling
  - Add all Pages accordingly
 - Create PathNotFoundPage for unmatched symbol
 - Login page now can log you in
   - Note: it does NOT check the username or password
   - It's literally just the onClick
 - HeaderBar can log you out
   - conditional rendering for headerbar button
   - HeaderBar tests moved to corresponding folder
 - Homepagecontents now has conditional rendering for admin panel
 - Rename path to from "/Streaming" to "/Streams" for StreamList page
 - Make all paths ONLY use `/Home` to go to dashboard
 - Clean up API calls to only happen at the level they are needed (no more endless chaining dataSource)
 - Tests updated
   - Login page new Auth functions being passed
   - Most tests that were being passed "datasource" props
   - Test added for index .jsx
Closes #338 

 
 
 
| Original   | Updated|
|:---------------:|:---------------:|
| | |

# Testing

**Test coverage**:

# Additional notes

**Note to reviewers**:

Don't forget to `npm i` before running this code. You need to grab the new library

**To do before merging**:

- [x]  Login Page Merge #362
- [x]  HomePage Merge #361 
- [x]   StreamListPage Merge #360 
- [x]  CreateStream Page Merge #363 
- [x]  DeviceLIst Page Merge #364 
- [x]  Logs Page Merge #365 
- [x]  DeviceDetails page Merge #366

PR #376, opened 2021-03-10
Rename path

jun-loke:
# General info

This PR only exists to make the #372  PR smaller. This is a change not related to the issue that the PR is supposed to be addressing and it can be easily reviewed on its own. SO i'm doing that
**Issue number**: 

**Task description**: 

 - Rename `/Streaming` path to `/Streams`
 
# Additional notes

**Note to reviewers**:


felixlapierre (approved):

gdzooks (approved):
system test needs a rename too but it's also gettng fixed up later so idm skippin out on it for now

Issue #330, opened 2021-02-22
Add protected routes to frontend


felixlapierre:
Redirect to login page when accessing a protected page and not logged in

PR #382, opened 2021-03-11
[Auth] Integrate protected routes with frontend auth 2

felixlapierre:
# General info
Integrate the protected routes code with frontend auth.

**Issue number**: Related to #330 

## Changelog

- Store JWT in cookies instead of local storage (because cookies expire automatically, while local storage we would have to manually write code to handle expiry)
- Add code to parse JWT in order to check expiry and user type

- Fixed an issue where redirecting from login to home refreshes for no reason
- Fixed an issue where logging out does not redirect to login
- Fixed an issue where CORS requests were disabled for DELETE method

- Moved some functions from AuthenticationApi to AuthenticationUtil
- Removed some functions from AuthenticationApi that are no longer needed
- Refactoring to account for moved and renamed functions

### DIFF
--- a/frontend/src/api/AuthenticationUtil.js
+++ b/frontend/src/api/AuthenticationUtil.js
@@ -7,7 +7,8 @@ function parseJwt(token) {
     atob(base64)
       .split("")
       .map((c) => {
-        return `%${`00${c.charCodeAt(0).toString(16)}`.slice(-2)}`;
+        const component = `00${c.charCodeAt(0).toString(16)}`;
+        return `%${component.slice(-2)}`;
       })
       .join("")
   );
--- a/frontend/src/api/AuthenticationUtil.js
+++ b/frontend/src/api/AuthenticationUtil.js
@@ -18,7 +18,7 @@ function parseJwt(token) {
 export function saveToken(token) {
   const tokenPayload = parseJwt(token.split(" ")[1]);
   const expiryDate = new Date(tokenPayload.exp * 1000);
-  Cookies.set("authToken", token, { expires: expiryDate });
+  Cookies.set("authToken", token, { expires: expiryDate, SameSite: "Strict" });
 }
 
 export function getAuthorizationHeader() {
--- a/frontend/src/api/AuthenticationUtil.js
+++ b/frontend/src/api/AuthenticationUtil.js
@@ -6,7 +6,7 @@ function parseJwt(token) {
   const jsonPayload = decodeURIComponent(
     atob(base64)
       .split("")
-      .map(function (c) {
+      .map((c) => {
         return `%${`00${c.charCodeAt(0).toString(16)}`.slice(-2)}`;
       })
       .join("")
--- a/frontend/src/api/AuthenticationUtil.js
+++ b/frontend/src/api/AuthenticationUtil.js
@@ -36,7 +36,7 @@ export function isAuthenticated() {
 export function isAdmin() {
   const token = Cookies.get("authToken");
   const tokenPayload = parseJwt(token.split(" ")[1]);
-  return tokenPayload.sub === "admin";
+  return tokenPayload.role === "ADMIN" || tokenPayload.role === "SUPERUSER";
 }
 
 export default {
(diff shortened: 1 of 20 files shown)


frontend/src/streamlist/DeleteStreamDialog.jsx
```diff
@@ -22,7 +22,7 @@ export default class DeleteStreamDialog extends React.Component {
 
   confirmDelete() {
     const { deleteId } = this.props;
-    deleteStream(deleteId).finally(this.afterDelete());
+    deleteStream(deleteId).then(this.afterDelete.bind(this));
```
jun-loke: Good catch! Didn't realize I was just calling the function. 
Though the binding should happen in the constructor (at time of creation)

jun-loke (changes requested):
Just a small tweak. Nothing major. Looks great

frontend/src/streamlist/DeleteStreamDialog.jsx
felixlapierre: Fixed in 61f245fb

jun-loke (dismissed):
all good now 😃

frontend/src/api/AuthenticationUtil.js
```diff
@@ -1,14 +1,44 @@
...
 
+export function isAuthenticated() {
+  return !!Cookies.get("authToken");
+}
+
+export function isAdmin() {
+  const token = Cookies.get("authToken");
+  const tokenPayload = parseJwt(token.split(" ")[1]);
+  return tokenPayload.sub === "admin";
+}
```
ButlerJAL: This is incorrect. `sub` is the username. You'll need to merge #354 which added roles. You can then do something like tokenPayload.role === "ADMIN"

ButlerJAL (commented):
The recommended way of storing JWT is in LocalStorage. Cookies are susceptible to CSRF attacks. Granted, the backend logic will ignore the cookie so the security risk is not really a factor here. I see that we are mostly using the cookie as a "auto-expiring" storing mechanism, so, I am not really sure how I feel about this meaning if it's easier leave it as is.

Else, I left a comment about checking roles such that we should merge #354 to do this.

MrJCipherButtles:
> The recommended way of storing JWT is in LocalStorage. Cookies are susceptible to CSRF attacks. Granted, the backend logic will ignore the cookie so the security risk is not really a factor here. I see that we are mostly using the cookie as a "auto-expiring" storing mechanism, so, I am not really sure how I feel about this meaning if it's easier leave it as is.
> 
> Else, I left a comment about checking roles such that we should merge #354 to do this.

Oh, was the backend logic changed in a previous PR regarding authentication? I just pulled the latest changes on master, and I seem to get errors?

felixlapierre:
@MrJCipherButtles I did some research about the pros and cons of putting JWT in cookies and localstorage. From what I can tell they both pose different security risks. Localstorage can be accessed by any javascript running on the page, so it's vulnerable to foreign scripts running in the page, while cookies are vulnerable to csrf attacks. For now I'm using cookies to store it to take advantage of the expiry mechanism, because either way there will be a security risk. If we find out that local storage is actually safer, we can switch to that really easily in a later PR.

frontend/src/api/AuthenticationUtil.js
felixlapierre: I'll fix this once #354 is merged

felixlapierre: Fixed in f14cc12

ButlerJAL (approved):
As implemented I do not believe one is safer than the other. LocalStorage is "safer" because you eliminate CSRF from the conversation. CSRF is only a conversation starter when there are Cookies.

XSS attacks are the more dangerous attack, but they can happen whether you use LocalStorage or Cookies. It's true that an XSS attack can bypass any CSRF countermeasures, but XSS countermeasures do not protect against CSRF attacks. It's important that the train of thought not be because XSS attacks exist that we do not care for CSRF attacks.

As implemented, the cookie is simply a storage device for the front-end logic; thus, our back-end service continues to be safe against CSRF attacks.

With that said, functionality proves to be great and I approve this PR 👍

gdzooks (approved):
looks clean to me! :tada:

PR #405, opened 2021-03-15
Refactor Tests to use toBe(true) instead of toBeTruthy

gdzooks:
# General info

**Issue number**: n/a

**Task description**: 
 - Updated tests using `toBeTruthy` to use `toBe(true)` instead since the latter is better practice & the former isn't necessary for the tests to work

# Testing

**Test coverage**:
- unchanged

# Additional notes
- mwah 😚


jun-loke (approved):
`expect(approved).toBe(true)`

sharon-ho (approved):

PR #411, opened 2021-03-18
[#401] Allow partial updates for encoders and decoders

### LINKED ISSUE #401: Allow partial updates for all PUT requests
Allow partial updates

mohhef:
# General info
allow partial updates for encoders and decoders


felixlapierre (approved):
Tested it locally to validate that it works. Not much to say here since it's a change we've made before a few times. LGTM!

ButlerJAL (approved):
LGTM

PR #412, opened 2021-03-19
[#315] Secrets

### LINKED ISSUE #315: Assign the superuser username, password and JWT Token secret in the CI/CD Pipeline
The superuser username, password, and JWT Token should be set from a GitHub Secret for security reasons. The can be changed by with the following spring properties:
security.authentication.jwt.secret
security.authentication.superuser.username
security.authentication.superuser.password

ButlerJAL:
Get secrets from System environment variables for Heroku continuous deployment.

# General info

**Issue number**: #315 

**Task description**: 

 - Programmatically set secrets to use System environment variables when default values are not to be used such as on Heroku deployment.
 - Added production secrets in Heroku.

# Testing

**Test coverage**: n/a

# Additional notes

**Note to reviewers**: n/a

**To do before merging**: n/a


src/main/resources/application.properties
```diff
@@ -4,8 +4,3 @@ spring.jpa.hibernate.ddl-auto=update
 spring.jpa.properties.hibernate.format_sql=true
 logging.level.org.hibernate.SQL=DEBUG
 logging.level.org.hibernate.type.descriptor.sql.BasicBinder=TRACE
-security.authentication.jwt.expirationtime=86400000
-security.authentication.jwt.secret=switchboardsecret
-security.authentication.superuser.username=admin
-security.authentication.superuser.password=admin
-security.authentication.url=/auth/token
```
FGRCL, line 11: You don't want to keep the authentication.url to be configurable?

src/main/resources/application.properties
```diff
@@ -4,8 +4,3 @@ spring.jpa.hibernate.ddl-auto=update
 spring.jpa.properties.hibernate.format_sql=true
 logging.level.org.hibernate.SQL=DEBUG
 logging.level.org.hibernate.type.descriptor.sql.BasicBinder=TRACE
-security.authentication.jwt.expirationtime=86400000
-security.authentication.jwt.secret=switchboardsecret
-security.authentication.superuser.username=admin
-security.authentication.superuser.password=admin
```
FGRCL, line 10: I think you might be able to do `${JWT_SECRET}` to get the environment variables. Then you could copy this file and name it `application-development.properties` so that we don't have to define those env variables for local development.

ButlerJAL: You don't need to set env variables for local development. Defaults will be used, if no env variables are set.

src/main/resources/application.properties
ButlerJAL: Configuration is done in SecurityProperties.java, no need to add them here.

ButlerJAL: Also "/auth/token" was never being used, so, not too sure what it was all about in the first place? The SecurityProperties was always using "/login" as the authentication url

FGRCL: Gotcha, I would change the value of AUTHORIZATION_URL in SecurityProperties.java to `/auth/token` otherwise it'll break the FE.

src/main/java/org/beanpod/switchboard/config/SecurityProperties.java
```diff
@@ -1,27 +1,24 @@
...
 
-  @Value("${security.authentication.jwt.secret}")
-  private String secret;
+  private final String superuserUsername =
+      System.getenv("SUPERUSER_USERNAME") != null ? System.getenv("SUPERUSER_USERNAME") : "admin";
 
-  @Value("${security.authentication.url}")
-  private String authenticationUrl = "/login";
+  private final String superuserPassword =
+      System.getenv("SUPERUSER_PASSWORD") != null ? System.getenv("SUPERUSER_PASSWORD") : "admin";
```
felixlapierre: I would prefix the environment variables with `SWITCHBOARD_` to avoid conflicts with other environment variables (though a conflict is unlikely to happen in practice, so feel free to ignore)

felixlapierre (dismissed):
Very minor suggestion, otherwise LGTM! Manually validated on my end that local usage works fine.

src/main/java/org/beanpod/switchboard/config/SecurityProperties.java
ButlerJAL: Done 👍

felixlapierre (approved):

FGRCL (approved):

PR #421, opened 2021-03-27
[#261] Allow devices to have configuration templates

### LINKED ISSUE #261: Store a configuration file for each device
After a discussion with the PM the requirements have shifted a bit:

- Store a file containing the configuration of each device (not each decoder). The format of the file is not assumed to be JSON, it could be any kind of file, and we will not perform any validation on it.
- We will not store a configurationSchema since we will allow configurations to have any file format. We will leave it up to the device to reject a configuration if it does not like it.
- Allow devices to register without providing a starting configuration.
- Allow users to call an endpoint to upload a new configuration file which will replace the old one.
- Allow users to call an endpoint to retrieve the current configuration
- Track the time at which the configuration was last modified. Make sure this value cannot be "manually" changed, it should be set to the current time whenever the config is changed.
- Store the config as a binary file in the db to disallow database injection.

mohhef:
# General info
1. The task is done on a separate endpoint because, in spring boot multipartfile, it is not allowed to upload a file + have a request body
2. I save files with bytes, thus should not be able to do db attacks

3. to test this pr
image


src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -46,6 +51,27 @@
   private final EncoderService encoderService;
   private final MaintainDeviceStatus maintainDeviceStatus;
 
+  @SneakyThrows
+  @PutMapping(
+      value = "config/{serialNumber}",
+      consumes = {MediaType.MULTIPART_FORM_DATA_VALUE})
+  public String uploadImage(
```
felixlapierre: Rename method

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
```diff
@@ -29,6 +33,22 @@
...
+      @PathVariable @Valid String serialNumber,
+      @RequestParam(value = "configurationInstance", required = false) MultipartFile instance) {
+
+    DeviceModel deviceModel = new DeviceModel();
+    deviceModel.setSerialNumber(serialNumber);
+    deviceModel.setConfigurationInstance(instance.getBytes());
+    deviceModel.setConfigurationLastModified(OffsetDateTime.now());
+
+    updateDevice(deviceModel);
+    return ResponseEntity.ok("Configurations uploaded.");
```
felixlapierre: Change to "Configuration uploaded." since there is only one file now.

src/test/java/org/beanpod/switchboard/controller/DeviceControllerTest.java
```diff
@@ -52,6 +53,19 @@ void setup() {
     MockitoAnnotations.initMocks(this);
   }
 
+  @Test
+  final void testUploadJson() {
```
felixlapierre: Rename, since we aren't doing exclusively json anymore

src/test/java/org/beanpod/switchboard/fixture/DeviceFixture.java
```diff
@@ -16,14 +18,18 @@
   public static final String DISPLAY_NAME = "Device #1";
   public static final String DISPLAY_NAME2 = "Device #2";
   public static final String STATUS = "offline";
+  public static final MockMultipartFile validJsonFile =
+      new MockMultipartFile("json", "", "application/json", "{\"json\": \"someValue\"}".getBytes());
```
felixlapierre: Rename, since the type doesn't have to be json anymore. Value is probably fine though.

src/test/java/org/beanpod/switchboard/controller/DeviceControllerTest.java
```diff
@@ -52,6 +53,19 @@ void setup() {
...
+
+    when(deviceService.findDevice(any())).thenReturn(Optional.of(deviceDTO));
+    when(deviceMapper.toDeviceDto((DeviceModel) any())).thenReturn(deviceDTO);
+    when(deviceService.save(any())).thenReturn(deviceDTO);
+    when(deviceMapper.toDeviceModel(any())).thenReturn(deviceModel);
+
+    String s1 =
+        deviceController.uploadConfiguration(DeviceFixture.SERIAL_NUMBER, validJsonFile).getBody();
+    assertEquals("Configurations uploaded.", s1);
```
felixlapierre: Should assert somewhere that the device being saved in `deviceService` actually has the correct `configurationInstance`

contracts/api-spec.yaml
```diff
@@ -151,6 +151,31 @@ paths:
...
+      tags:
+        - Devices
+      requestBody:
+        content:
+          multipart/form-data:
+            schema:
+              properties:
+                configurationInstance:
+                  type: string
+                  format: binary
```
felixlapierre: We should just rename this to `configuration` since we don't need a distinction between instance and schema anymore.

felixlapierre (commented):
Mostly suggesting renames here. Also, we should probably wait until restrict access is merged before merging this.

contracts/api-spec.yaml
mohhef: fixed

src/test/java/org/beanpod/switchboard/fixture/DeviceFixture.java
mohhef: fixed

src/test/java/org/beanpod/switchboard/controller/DeviceControllerTest.java
mohhef: fixed

src/main/java/org/beanpod/switchboard/controller/DeviceController.java
mohhef: fixed

src/test/java/org/beanpod/switchboard/controller/DeviceControllerTest.java
mohhef: how would you verify that?

felixlapierre:
```
DeviceDto expectedSavedDevice = DeviceFixture.getDeviceDto();
expectedSavedDevice.setConfigurationInstance(validJsonFile.getBytes());
verify(deviceService).save(expectedSavedDevice);
```

felixlapierre (dismissed):

src/test/java/org/beanpod/switchboard/controller/DeviceControllerTest.java
mohhef: Fixed

sebhani (dismissed):
Works fine!

felixlapierre (approved):

sebhani (approved):

PR #422, opened 2021-03-27
[#259] Add snackbar messages

### LINKED ISSUE #259: Toast messages should display when any user action is performed (i.e. stream and delete)
After a stream or device is deleted and after a stream is initiated, a toast message should appear on the frontend to let the user now the action has been performed.

sharon-ho:
# General info
**Issue number**: Closes #259

**Task description**: 
 - Snackbar component resides in `AppRouter`
 - Anytime a snackbar should be called, the user should import the function `snackbar` from `SnackbarMessage.jsx` and call the function with parameters `status` (i.e. "success", "error") and `message` (string)

# Testing

**Test coverage**: 97.4%

# Additional notes

**Note to reviewers**:
- It takes a while for the messages to appear if backend is not setup.
- Test for `DeleteDeviceButton` is considered not up to standard with our current testing methodology. However the scope of my PR is not to fix this test, hence I am only adding what is necessary after the implementation of the snackbar to this existing test. On top of the fact that we might remove that button entirely very soon to replace the action menu in the device table with the device details icon button, similar to the one in the stream list.

**To do before merging**:
- N/A


frontend/package.json
```diff
@@ -19,6 +19,7 @@
     "prop-types": "^15.7.2",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
+    "react-router": "^5.2.0",
```
felixlapierre: Were we not already using react-router? Also I'm surprised this didn't change package-lock

frontend/src/general/SnackbarMessage.jsx
```diff
@@ -0,0 +1,166 @@
...
+      this.setIsSuccess(true);
+    } else {
+      this.setIsSuccess(false);
+    }
+    this.setMessage(msg);
+  }
+
+  openSnackbar(stat, msg, path) {
+    const currentPathname = this.setPathname(path);
+    if (currentPathname !== "") {
```
felixlapierre: Why make `this.setPathname` return the pathname? I'd suggest instead doing
```
this.setPathname(path);
if (path !== "") {
```

frontend/src/general/SnackbarMessage.jsx
```diff
@@ -0,0 +1,166 @@
...
+    return pathname;
+  }
+
+  addSnackbarProperties(stat, msg) {
+    this.setStatus(stat);
+    if (stat !== "error") {
+      this.setIsSuccess(true);
+    } else {
+      this.setIsSuccess(false);
+    }
```
felixlapierre: Simplify to
```
this.setIsSuccess(stat !== "error")
```

frontend/src/general/SnackbarMessage.jsx
```diff
@@ -0,0 +1,166 @@
...
+
+  refresh() {
+    const { history, location } = this.props;
+    const { pathname } = this.state;
+    if (location.pathname.endsWith(pathname)) {
+      history.go(0);
+    } else {
+      history.push(`/${pathname}`);
+    }
+  }
```
felixlapierre: I feel like refreshing and redirecting should not be a responsibility of the snackbar message. The snackbar message should probably only be responsible for displaying the message itself.

With the current design there is a lot of conditional logic here - if pathname is empty do nothing, if it's nonempty but we're on that page then refresh, if it's nonempty and we're not on that page redirect... I feel like it would be simpler if the component that opens the snackbar is the one who has to call `history.push` or `history.go` if they want to.

frontend/src/general/__tests__/DeleteDeviceButton.test.jsx
```diff
@@ -91,64 +79,60 @@ describe("DeleteButton", () => {
...
 
-          // click confirm
-          wrapper.find("#confirmDeleteBtn").simulate("click");
+        // click confirm
+        wrapper.find("#confirmDeleteBtn").simulate("click");
 
-          // check that axios.delete was called
-          expect(axios.delete).toHaveBeenCalled();
+        // check that axios.delete was called
+        expect(axios.delete).toHaveBeenCalled();
```
felixlapierre: Comments like this are kind of redundant imo, i'd remove them. The whole point of jest's `expect` is that you can read the assert line like english "expect axios.delete to have been called" so they don't need a comment explaining what they're doing.

frontend/src/general/SnackbarMessage.jsx
```diff
@@ -0,0 +1,166 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Snackbar, SnackbarContent, IconButton, Box } from "@material-ui/core";
+import { CheckCircle, Error, Close } from "@material-ui/icons";
+import { withRouter } from "react-router";
+
+let openSnackbarFn;
```
felixlapierre: You can rename this to openSnackbar without getting name conflicts since the other openSnackbar is a member function of the SnackbarMessage class.

frontend/src/general/SnackbarMessage.jsx
```diff
@@ -0,0 +1,166 @@
...
+    go: PropTypes.func.isRequired
+  }).isRequired,
+  location: PropTypes.shape({
+    pathname: PropTypes.string.isRequired
+  }).isRequired
+};
+
+export function snackbar(status, message, pathname = "") {
+  openSnackbarFn(status, message, pathname);
+}
```
felixlapierre: So from what I can tell, this is how snackbar works

- One SnackbarMessage component exists in AppRouter (so it exists in all pages)
- Any component can import this `snackbar` global function
- When calling `snackbar` it opens the snackbar from anywhere, with the given status and message

Usually I would say exposing a global function like this is kind of an antipattern, but I guess the only other way to do it is to pass the snackbar function along in all the props of every component, which would be less clean.

My question is: is this the "standard" for how snackbar opening is done? Do you have an article or something that you followed for this design? I think it would be a useful reference for me to trust that this is a tried-and-true architecture decision.

frontend/src/general/SnackbarMessage.jsx
```diff
@@ -0,0 +1,166 @@
...
+          }}
+          open={open}
+          autoHideDuration={3000}
+          onClose={this.handleClose}
+        >
+          <SnackbarContent
+            style={{
+              backgroundColor: isSuccess ? "#4caf50" : "#f44336"
+            }}
+            // prettier-ignore
```
felixlapierre, line 60: Why `prettier-ignore`?

frontend/src/general/SnackbarMessage.jsx
```diff
@@ -0,0 +1,166 @@
...
+            message={(
+              <Box id='message-id' className="snackMessage">
+                {isSuccess ? 
+                  <CheckCircle className="iconPadding" /> : 
+                  <Error className="iconPadding" />}
+                {message || `Form submission status: ${status}`}
```
felixlapierre: From what I can tell we always pass the message, so I think we can remove the `Form submission status` case here.

frontend/src/general/SnackbarMessage.jsx
```diff
@@ -0,0 +1,166 @@
...
+    const { pathname } = this.state;
+    if (location.pathname.endsWith(pathname)) {
+      history.go(0);
+    } else {
+      history.push(`/${pathname}`);
+    }
+  }
+
+  render() {
+    const { open, isSuccess, status, message } = this.state;
```
felixlapierre: I think you can get rid of either `status` or `isSuccess` from the state.

You could remove `isSuccess` from the state and just put this line here:
```
const isSuccess = status === "success";
```

Or you could remove status entirely and have the called pass `isSuccess` directly, since the only place you use the status value itself is when the message is missing, and the message is never missing. (personally I prefer this option because it means callers don't need to know the possible values for `status`)

frontend/src/general/SnackbarMessage.jsx
```diff
@@ -0,0 +1,166 @@
...
+
+  componentDidMount() {
+    openSnackbarFn = this.openSnackbar;
+  }
+
+  handleClose(event, reason) {
+    if (reason === "clickaway") {
+      return;
+    }
+    this.setOpen(false);
```
felixlapierre: Simplify with
```
if (reason !== "clickaway") {
  this.setOpen(false);
}
```

felixlapierre (changes requested):
I have some suggestions and questions about the SnackbarMessage! Let me know if anything is unclear.

frontend/package.json
sharon-ho: We can remove this. I believe react-router and react-router-dom point to the same package.

frontend/src/general/SnackbarMessage.jsx
sharon-ho: This is because if we do this the `path != ""` will be evaluated at the same time as the `setPathname()` is being called and thus will mess up with the if statement evaluation.

I tried doing the code you suggested early on and that's what happened. The only way I could bypass it from the couple of things I tried was making sure the setPathname was finished calling by returning a value and that value (which is equivalent to the pathname state that's just been set) would be the one to be evaluated in the if statement.

frontend/src/general/SnackbarMessage.jsx
sharon-ho: This would lead to a lot of props passing of the useHistory() hook, but it could be possible to wrap all the class components (which is the primary react component type we are using for the project) with the withRouter function from react-router-dom. I wasn't sure if that was better architecturally to do so.

frontend/src/general/__tests__/DeleteDeviceButton.test.jsx
sharon-ho: I added the comment cause some of the tests had comments like this, but I'll remove it!

frontend/src/general/SnackbarMessage.jsx
sharon-ho: I was inspired by this guide written by freecodecamp.

https://www.freecodecamp.org/news/how-to-show-informational-messages-using-material-ui-in-a-react-web-app-5b108178608/

https://github.com/async-labs/builderbook/blob/master/book/4-end/components/Notifier.jsx

frontend/src/general/SnackbarMessage.jsx
sharon-ho: This is because the linter will make the following couple lines with the conditional rendering not very "human-readable" with how they indent and space out the next couple of lines, but I can remove it if that's an issue.

frontend/src/general/SnackbarMessage.jsx
sharon-ho: I opted not to do this because it is convention to write the handleClose() function in this way:
image

Lines 28-34 from this material-ui example code: https://github.com/mui-org/material-ui/blob/master/docs/src/pages/components/snackbars/CustomizedSnackbars.js

frontend/src/general/SnackbarMessage.jsx
sharon-ho: Also with the way the snackbar is architectured, I think it would be difficult to have the component handle the refreshing and then calling the snackbar. Originally, before I changed snackbar to appear after refresh, the refresh would be dependent on the snackbar's close button and autoHide. However I'm wondering if the refresh might unmount the component that called the refresh, and with the way snackbar is architectured, the component might not even call the snackbar exported function after a refresh if the component that got unmounted is no longer existent. But I haven't actually investigated to see if that would happen.

frontend/src/general/SnackbarMessage.jsx
felixlapierre: I feel like that's less a convention and more of the way that guy wrote it. I would still be in favor of changing it to my suggestion provided it still works.

frontend/src/general/SnackbarMessage.jsx
felixlapierre: I don't really mind that much

frontend/src/general/SnackbarMessage.jsx
felixlapierre: Thank you!!

frontend/src/general/SnackbarMessage.jsx
sharon-ho: After some testing with @felixlapierre we discovered that the comment above can no longer be replicated, so the fix that was suggested will be implemented.

frontend/src/general/SnackbarMessage.jsx
sharon-ho: After some testing with @felixlapierre we discovered that the refresh can be removed as a responsibility of the snackbar, so the fix that was suggested will be implemented.

frontend/package.json
sharon-ho: Actually I think react-router-dom extends from react-router, which was already in the package-lock.

frontend/src/general/SnackbarMessage.jsx
sharon-ho: Will get rid of isSuccess as a state after seeing your other comment about setting it as a constant dependent on the status state in `render()`.

felixlapierre:
From a functionality point of view, I've tested it and it works great :smile:

frontend/src/general/__tests__/SnackbarMessage.test.jsx
```diff
@@ -0,0 +1,152 @@
...
+        expect(wrapper.find(SnackbarContent)).toHaveLength(1);
+      });
+
+      describe("The <SnackbarContent/> component", () => {
+        describe("has a style prop", () => {
+          const statusColors = {
+            green: "#4caf50",
+            red: "#f44336"
+          };
+          describe("where if status", () => {
```
felixlapierre: I suggest reorganizing this into
````
Describe SnackbarContent
    If status is success
        Has a green background and the correct props
        Contains one <CheckCircle/> component
    If status is not success
        Has a red background and the correct props
        Contains one <Error/> component
````

File: frontend/src/general/__tests__/SnackbarMessage.test.jsx [hunk 13]
```diff
@@ -0,0 +1,152 @@
...
+              wrapper.setState({ status: "success" });
+              expect(wrapper.find(SnackbarContent).props().message).toEqual(
+                <Box id="message-id" className="snackMessage">
+                  <CheckCircle className="iconPadding" />
+                  {wrapper.state().message}
+                </Box>
+              );
+            });
+
+            it("is anything else, it should have an <Error/> icon component", () => {
```
felixlapierre: This phrasing makes it a bit disconnected and confusing - the subject of the sentence is in the previous block making the test name harder to read. Instead write "If status is ...."

Also, saying "anything else" is hard to understand without looking at the previous block. Instead write "If status is not success"

(I already suggested my preferred organizing for tests, just explicitly mentioning this here as a guideline for future tests)

frontend/src/general/__tests__/DeleteDeviceButton.test.jsx
```diff
@@ -91,65 +94,67 @@ describe("DeleteButton", () => {
...
 
-          // check that redirect has been called
           expect(mockHistoryPush).toHaveBeenCalledWith("/Devices");
-
-          // open should be false
-          expect(setOpen).toHaveBeenCalledWith(false);
         });
       });
+      it("If the axios.delete is rejected, it should close the dialog and display an error snackbar", async () => {
+        wrapper.find("#deleteBtn").simulate("click");
```
felixlapierre, line 138: Pretty sure we don't use `.simulate("click")` anymore, but I forget what the alternative is when it's a functional component and we can't directly call the function. (I might be mistaken tho)

felixlapierre (dismissed):
Only very minor comments about a few tests. GREAT WORK SHARON!!!!

frontend/src/general/__tests__/DeleteDeviceButton.test.jsx
sharon-ho: I mentioned in the PR notes that I'm not going to redo the test to fit with our current testing methodology since it's not in the scope of the PR. On top of the fact that we might remove this button entirely very soon to replace the action menu in the device table with the device details icon button, similar to the one in the stream list.

frontend/src/general/__tests__/SnackbarMessage.test.jsx
```diff
@@ -0,0 +1,156 @@
...
+        };
+        describe("If status is success,", () => {
+          describe("the snackbar has a green background and the correct props", () => {
+            it("and contains one <CheckCircle/> component", () => {
+              wrapper.setState({ status: "success" });
+              expect(wrapper.find(SnackbarContent).props().style).toEqual({
+                backgroundColor: statusColors.green
+              });
+            });
+          });
```
felixlapierre: I think you got mixed up here - your `it` mentions the CheckCircle component but doesn't actually check it's there.
Also you have two nested describes for a single test. The purpose of describe is to separate tests into groups, so having 2 of them surrounding one test is kind of missing the point.

You're almost there though! Try organizing the tests like this

```
describe("<SnackbarContent/> component")
    describe("If status is success")
        it("has a green background")
        it("contains a message prop with a <CheckCircle/> component")
    describe("If status is not success")
        it("Has a red background")
        it("Contains a message prop with a <Error/> component")
```

felixlapierre (commented):
You missed the mark a bit responding to one of my comments - just a bit more!

frontend/src/general/__tests__/SnackbarMessage.test.jsx
sharon-ho: Oh okay, I think I misunderstood your previous comments. Thanks for clarifying! I'll update accordingly.

frontend/src/general/__tests__/SnackbarMessage.test.jsx
```diff
@@ -0,0 +1,146 @@
...
+  it,
+  jest
+} from "@jest/globals";
+import { Snackbar, SnackbarContent, Box } from "@material-ui/core";
+import { CheckCircle, Error } from "@material-ui/icons";
+import SnackbarMessage from "../SnackbarMessage";
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe("<SnackbarMessage /> Class Component", () => {
```
gdzooks, line 18: i see the exported `snackbar()` isn't tested in here. i know it gets tested in other test files but sonarcloud still shows the line as uncovered. is it not possible to just call `snackbar()` in here to check that it's calling `openSnackbar()` with the correct info?

gdzooks (approved):
one comment but this has been up for so long & i've helped test it a bunch so... it pretty much works as we want it to and while the coverage isn't the easy 100% we've been getting on frontend PRs it's over 95% so i'm willing to approve as-is!

frontend/src/general/__tests__/SnackbarMessage.test.jsx
sharon-ho: I believe it's more complicated than that and I personally made the decision to not spend additional time investigating how to test it.

felixlapierre (approved):
Comments addressed, LGTM!

PR #428, opened 2021-03-29
[#328] Retrieve specific stream logs

### LINKED ISSUE #328: Endpoint to get logs related to a stream
Such logs include
- log that is created when the stream is started
- logs of devices reporting errors during the stream

sebhani:
# General info
Add an endpoint to retrieve X stream logs.

**Issue number**: #328 - the issue contains another task that will be handled separately from this.

**Task description**: 

 - Add a controller for the /streamlogs/{streamId} endpoint where it returns all the logs for that specific stream.

# Additional notes

**Note to reviewers**:
Create and delete multiple streams and then try to retrieve the logs of a specific one of them.


src/main/java/org/beanpod/switchboard/controller/StreamLogController.java
```diff
@@ -0,0 +1,21 @@
...
+@RestController
+@RequiredArgsConstructor
+public class StreamLogController implements StreamlogApi {
+  private final StreamLogDaoImpl streamLogDao;
+
+  @Override
+  public ResponseEntity<List<StreamLogModel>> retrieveStreamLogs(@PathVariable Long streamId) {
+    return ResponseEntity.ok(streamLogDao.getStreamLogs(streamId));
+  }
+}
```
mohhef: why not make the stream logs in the logs controller

mohhef: also maybe the endpoint should start with /log to  be consistent with the log controller

mohhef (changes requested):
Looks good to me overall, tested works fine, i have one comment.
I think we should keep getting stream logs in the logs controller

src/main/java/org/beanpod/switchboard/controller/StreamLogController.java
sebhani: I was going to do it this way, but /log/{XYZ} is already taken by another log endpoint. I can do something like /log/stream/{streamID}, what do you think?

mohhef: yeah i was thinking /log/stream/{streamID}

src/test/java/org/beanpod/switchboard/dao/StreamLogDaoImplTest.java
```diff
@@ -41,4 +47,14 @@ final void createStreamLogTest() {
 
     assertEquals(streamLogDto, actualStreamLog);
   }
+
+  @Test
+  final void getStreamLogsTest() {
+    when(logStreamMapper.toStreamLogModels(any())).thenReturn(streamLogModelList);
+    when(logStreamRepository.findByStreamId("1")).thenReturn(streamLogList);
+
+    List<StreamLogModel> listOfStreamLogModels_actual = streamLogDao.getStreamLogs((long) 1);
```
felixlapierre: Isn't there a way to do it without casting? Like `1L` instead of `(long) 1`? Or does that not work?

src/main/java/org/beanpod/switchboard/controller/StreamLogController.java
felixlapierre: I agree with @mohhef

src/test/java/org/beanpod/switchboard/controller/StreamLogControllerTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.openapitools.model.StreamLogModel;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+
+public class StreamLogControllerTest {
```
felixlapierre: I agree that this could be part of `LogController`. Or would that be too many responsibilities in one class? It would make sense though if we change the endpoint to `/log/stream/{id}`.

felixlapierre (dismissed):
Only minor suggestions. I'll re-approve once you make the change requested by mohamed.

src/main/java/org/beanpod/switchboard/controller/StreamLogController.java
sebhani: the endpoint has been updated to /log/stream/{streamID}

src/test/java/org/beanpod/switchboard/dao/StreamLogDaoImplTest.java
sebhani: Thanks for the suggestion. I forgot that we can do something like #L to represent long number :)! I have updated my tests accordingly.

felixlapierre (approved):
My changes were addressed, LGTM!

mohhef (approved):
i might have also moved the streamlogdao to the logdao but its fine i think

PR #430, opened 2021-03-30
[#271] Stream Statistics

gdzooks:
# General info

**Issue number**: closes #271

**Task description**: 
 - Fill in card for Stream Statistics on the Stream Details page
 - Add button to navigate to our very fancy new Detailed Stream Statistics Page
 - Add a Detailed Stream Statistics Page
 
 
| Original   | Updated|
|:---------------:|:---------------:|
| image | image |
| N/A | image |

# Testing

**Test coverage**:

# Additional notes

**Note to reviewers**:
- You need to add stats manually to your stream with Postman or something
- or sampleApp I guess xd

**To do before merging**:
- [x] Merge #431 + Fix conflicts
- [x] Merge #432

PR #434, opened 2021-03-31
Refactor add small dashboard card button

jun-loke:
# General info

**Issue number**: N/A

**Task description**: 

 - Refactor DashBoardCard to have Grid container around children
   - adapt all HOC that use DashboardCards accordingly and update their tests 
 - Create SmallCardButton and ButtonInfo objects for the dashabord card
 - Added conditional rendering for SmallCardButton using new optional button prop

 
 
| Original   | Updated|
|n/a |image
 |

# Testing

**Test coverage**:

# Additional notes

**Note to reviewers**:

To try this out, add the following to a card. I used the cards in StreamDetailsPageContents, since that's where these will be used, though these components are undergoing changes in #430 so the buttonCard will be implemented there or afterwards
``` javascript
+ import ButtonInfo from "../general/dashboard/ButtonInfo";
```
``` javascript
  const button = new ButtonInfo(
    `somePathName/${someDeviceInfo.serialNumber}`,
    {device: someDeviceInfo},
    "buttonTitle"
  );

<DashboardCard title="someTitle" button={button}>
```

**To do before merging**:

### DIFF
--- a/frontend/src/deviceDetailsPage/cards/DeviceChannelCard.jsx
+++ b/frontend/src/deviceDetailsPage/cards/DeviceChannelCard.jsx
@@ -13,10 +13,8 @@ export default function DeviceChannelCard(props) {
 
   return (
     <DashboardCard title="Channels">
-      <Grid container justify="center" direction="row" spacing={3}>
-        <Grid item xs={12}>
-          <ChannelDetailsTable channels={channels} />
-        </Grid>
+      <Grid item xs={12}>
+        <ChannelDetailsTable channels={channels} />
       </Grid>
     </DashboardCard>
   );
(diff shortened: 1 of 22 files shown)


felixlapierre:
Tried clicking the view device button, got a `TypeError: channels is undefined` error. 🍍

gdzooks (commented):
code looks good to me but unfortunately the button doesn't seem to be working on my end? i get this:
image

looking at how you coded the button i feel like it SHOULD work?
~~if you could do the receiver card as well while you're here that'd be pretty cash money-~~

felixlapierre (approved):
Im cool w/ merging this! Tho we should merge with the view device button on both the sender and receiver cards (or on neither)

gdzooks (approved):
same comment as felix! worst case if u dont add the button here i'll make myself a note to add it when i fix up #430 :3

jun-loke:
@felixlapierre  @gdzooks 

1) The issue with cannot read property map of undefined is due to issue where the DeviceInfo being retrieved from the StreamInfo does not have all of the properties defined. This has been logged in #437 Pr will be up soon

2) @gdzooks  Since you're replacing the deviceCards on Stream details in #430 I'll leave that up to you

PR #448, opened 2021-04-04
[#443] Fix Device Stealing

### LINKED ISSUE #443: User can create a device with the same serial number as a device belonging to another user
Steps to reproduce:
- Start the backend
- Create a second user
- As the default superuser A, register a device with a given serial number
- As the second user B, register a device with the same serial number

Expected behaviour:
- An error is returned saying that a device with that serial number already exists

Actual behaviour:
- Registration succeeds and user A no longer has access to the device with that serial number

ButlerJAL:
# General info

The create device action looks up the entire list of devices rather than the user's list of devices to avoid "device stealing".

**Issue number**: #443

**Task description**: 

 - Update createDevice method in DeviceController to lookup entire list of devices as to not override existing devices that belong to another user.

# Testing

**Test coverage**:
No change.

# Additional notes

**Note to reviewers**: n/a

**To do before merging**: n/a

### DIFF
--- a/src/main/java/org/beanpod/switchboard/controller/DeviceController.java
+++ b/src/main/java/org/beanpod/switchboard/controller/DeviceController.java
@@ -64,7 +64,7 @@ public class DeviceController implements DeviceApi {
     String publicIpAddress = request.getRemoteAddr();
 
     Optional<DeviceDto> deviceLookup =
-        deviceDao.findDevice(user, createDeviceRequest.getSerialNumber());
+        deviceDao.findDevice(createDeviceRequest.getSerialNumber());
     if (deviceLookup.isPresent()) {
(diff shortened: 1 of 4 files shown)


felixlapierre (approved):
I have validated that this works, LGTM!!

mohhef:
is that basically disallowing diffirent users to create devices with the same serial numbers?

ButlerJAL:
> is that basically disallowing diffirent users to create devices with the same serial numbers?

Yes. What would happen in the case where a user tried to create a device that belonged to another user is that the save transaction would go through and update the device. When creating devices, it’s important to lookup a device’s existence against all devices rather than the subset of user devices.

FGRCL (approved):
I tried it, hits the nail on the head.

PR #457, opened 2021-04-06
[#328] Endpoint to report stream errors

### LINKED ISSUE #328: Endpoint to get logs related to a stream

sebhani:
# General info
Created an endpoint that allows for creating stream-related logs. The new endpoint is /log/stream

**Issue number**: #328 

**Task description**: 

 - Update api-spec
 - Implement createStreamLog controller
 - Implement createLog method which is a wrapper method that accepts CreateStreamLogRequest object
 - Add tests

# Testing
To ease your testing process, here's the request that you need to ensure that it works:
Request to /log/stream endpoint

### DIFF
--- a/contracts/api-spec.yaml
+++ b/contracts/api-spec.yaml
@@ -579,12 +579,6 @@ components:
           type: string
           pattern: '^\d{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T(0[0-9]|1[0-9]|2[0-4])(\+00:00)$'
           example: '2021-02-04T08:21:23+00:00'
-        decoderSerial:
-          type: string
-          example: '1xw1'
-        encoderSerial:
-          type: string
-          example: '2Xe2'
 
     StreamLogModel:
       description: stream log object
--- a/contracts/api-spec.yaml
+++ b/contracts/api-spec.yaml
@@ -37,7 +37,23 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/StreamLogsModel'
-
+  /log/stream:
+    post:
+      operationId: createStreamLog
+      description: Create a stream log
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/createStreamLogRequest'
+      responses:
+        '200':
+          description: Successful creation of a stream log
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/StreamLogModel'
   /user/sign-up:
     post:
       operationId: sign-up
@@ -549,6 +565,26 @@ components:
         serialNumber:
           type: string
           example: '1xw1'
+
+    createStreamLogRequest:
+      description: A request to create a new stream log
+      properties:
+        streamId:
+          type: integer
+          example: '123'
+        message:
+          type: string
+          example: 'A log message'
+        dateTime:
+          type: string
+          pattern: '^\d{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T(0[0-9]|1[0-9]|2[0-4])(\+00:00)$'
+          example: '2021-02-04T08:21:23+00:00'
+        decoderSerial:
+          type: string
+          example: '1xw1'
+        encoderSerial:
+          type: string
+          example: '2Xe2'
 
     StreamLogModel:
       description: stream log object
--- a/src/main/java/org/beanpod/switchboard/controller/LogController.java
+++ b/src/main/java/org/beanpod/switchboard/controller/LogController.java
@@ -7,10 +7,13 @@ import lombok.RequiredArgsConstructor;
 import org.beanpod.switchboard.dao.LogDaoImpl;
 import org.beanpod.switchboard.dao.StreamLogDaoImpl;
 import org.beanpod.switchboard.dto.mapper.LogMapper;
+import org.beanpod.switchboard.dto.mapper.LogStreamMapper;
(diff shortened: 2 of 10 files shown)


contracts/api-spec.yaml
```diff
@@ -550,6 +566,26 @@ components:
...
+        dateTime:
+          type: string
+          pattern: '^\d{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T(0[0-9]|1[0-9]|2[0-4])(\+00:00)$'
+          example: '2021-02-04T08:21:23+00:00'
+        decoderSerial:
+          type: string
+          example: '1xw1'
+        encoderSerial:
+          type: string
+          example: '2Xe2'
```
felixlapierre: Would it be possible to remove these fields from the create stream request? Since we have the stream ID we can determine the encoder and decoder serial ourselves in the backend, we don't need to make the devices responsible for providing the serial number of the other device.

src/test/java/org/beanpod/switchboard/service/StreamLogServiceTest.java
```diff
@@ -47,4 +52,15 @@ final void createStreamLogTest() {
 
     assertEquals(streamLogDto, actualStreamLogDto);
   }
+
+  @Test
+  final void CreateStreamLogCreateStreamLogRequestTest() {
```
felixlapierre: I'm a bit confused by this test name, maybe `CreateStreamLogFromRequestTest`? But maybe there's something better.

src/main/java/org/beanpod/switchboard/controller/LogController.java
```diff
@@ -51,4 +56,14 @@
         .map(ResponseEntity::ok)
         .orElseThrow(() -> new ExceptionType.UnknownException(CONTROLLER_NAME));
   }
+
+  @Override
+  public ResponseEntity<StreamLogModel> createStreamLog(
+      CreateStreamLogRequest createStreamLogRequest) {
+    return Optional.of(createStreamLogRequest)
+        .map(streamLogService::createLog)
+        .map(logStreamMapper::toStreamLogModel)
```
felixlapierre: I just realized we should rename `LogStreamMapper` to `StreamLogMapper`

felixlapierre (changes requested):

src/test/java/org/beanpod/switchboard/service/StreamLogServiceTest.java
sebhani: This's what happens when you write tests before going to bed :)

src/main/java/org/beanpod/switchboard/controller/LogController.java
sebhani: Good catch! I rename the class

felixlapierre (approved):
Comments addressed, LGTM! Works great on my end too.

src/main/java/org/beanpod/switchboard/service/StreamLogService.java
```diff
@@ -27,8 +34,25 @@ public StreamLogDto createLog(
...
   }
+
+  public StreamLogDto createLog(CreateStreamLogRequest createStreamLogRequest) {
+    Optional<StreamDto> streamDto =
+        Optional.of(Long.valueOf(createStreamLogRequest.getStreamId()))
+            .map(streamDao::getStreamById);
+
+    if (!streamDto.isPresent()) {
+      return null;
+    }
```
ButlerJAL, line 49: Just as a tip, try to avoid using negation (!) when possible. Here you can use `.isEmpty()`. It's something small, but it helps with readability.

src/main/java/org/beanpod/switchboard/service/StreamLogService.java
```diff
@@ -27,8 +34,25 @@ public StreamLogDto createLog(
...
+
+    return createLog(
+        logMapper.map(createStreamLogRequest.getDateTime()),
+        createStreamLogRequest.getMessage(),
+        streamDto.get().getInputChannel().getDecoder().getSerialNumber(),
+        streamDto.get().getOutputChannel().getEncoder().getSerialNumber(),
+        createStreamLogRequest.getStreamId().toString());
+  }
```
ButlerJAL, line 57: You may use functional style as suggested below to make your method slightly more compact and more uniform rather than breaking it up with a if-statement.
```
public StreamLogDto createLog(CreateStreamLogRequest createStreamLogRequest) {
    return Optional.of(Long.valueOf(createStreamLogRequest.getStreamId()))
        .map(streamDao::getStreamById)
        .map(
            dto ->
                createLog(
                    logMapper.map(createStreamLogRequest.getDateTime()),
                    createStreamLogRequest.getMessage(),
                    dto.getInputChannel().getDecoder().getSerialNumber(),
                    dto.getOutputChannel().getEncoder().getSerialNumber(),
                    createStreamLogRequest.getStreamId().toString()))
        .orElse(null);
  }
```

ButlerJAL (approved):
Approved, LGTM! I left two possible tips for you to digest on, else things look good. I know you're still working on the code for logs, so you may choose to action one of the tips in your next PR 👍

src/main/java/org/beanpod/switchboard/service/StreamLogService.java
sebhani: Thanks for the suggestions. I'll make sure to apply them before merging next PR

PR #473, opened 2021-04-09
[#444] Make the Encoder Controller return api spec models

### LINKED ISSUE #444: All controllers should return objects using Model classes rather than objects using DTO classes

FGRCL:
# General info

This should fix the encoders returning the user's password hash

**Issue number**: #444 
**Task description**: 

 - Rewrite api spec
 - Implement the api spec in the Encoder Controller
 - Adapt the unit tests
 
# Testing

**Test coverage**:
Rewrote unit tests

# Additional notes

### DIFF
--- a/contracts/api-spec.yaml
+++ b/contracts/api-spec.yaml
@@ -244,7 +244,7 @@ paths:
 
   /encoder/{serialNumber}:
     get:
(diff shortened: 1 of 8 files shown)


FGRCL:
PR made from a fork

PR #450, opened 2021-04-05
[#445] Restrict Access to Streams

ButlerJAL:
# General info

I'm starting this PR as a draft to invite collaboration. Contents of PR are functional. Help with fixing broken tests (tests annotated by `@Disabled`) is always appreciated.

**Issue number**: : #445 

**Task description**: 

 - Restrict access for Stream
 - Fix Stream tests
 - Restict access for InputChannel and OutputChannel
 - Standardized Repository methods
 - Standardized Mapper methods

# Testing

**Test coverage**: No change

# Additional notes

**Note to reviewers**:
~Collaboration over review while this PR is in draft. Pushing additions/changes directly to this PR is allowed and welcomed.~

**To do before merging**: n/a

PR #474, opened 2021-04-09
[#446] Retrieve logs that belong to the signed-in user

### LINKED ISSUE #446: Restrict access to logs
When getting all logs, a user should only see the logs that belong to them.

sebhani:
# General info
Retrieve logs that only belong to the signed-in user

**Issue number**: #446 

**Task description**: 

 - Update the LogRepository class to retrieve logs that are owned by the specific user using a native query
 - Update the LogDaoImpl class to pass the signed-in user id
 - Update LogDaoImpl tests
 - Same updates as the above ones are done for StreamLog

# Testing
Make sure that the logs that are retrieved by /log and /log/{serialNumber} belong to the user (user who owns the device related to the log)

**Note**: ~~This task is the first part of #446 user story. The second part is for restricting access of stream-related logs~~ All parts has been done in this PR


ButlerJAL (commented):
@sebhani I had a look at your changes. I'd like to see the stream log changes in conjunction with these changes. So, I'd prefer if you continued adding to this PR to make one PR rather than splitting the log task up. Also, I have a few details I'd like to go over.

I'd "bubble up" the request handling logic such as getting the user from the security context up to the controller. The controller is responsible for handling inputs related to the request. The security context of the request being one of those inputs.

Ideally, try to keep the DAO as light as possible such that it should simply provide a connection to the DB (Repository). However, our code-base lacks a little standard regarding the Service Layer and the DAO Layer depending on who did what. Not everyone is comfortable with layer separation. Personally, where I found that the Service layer was cut out to make the DAO a little heavier weight or the Service layer skipped over the DAO, I did not bother much about it. You can make your own judgment call on what you see.

For the Repository, look at leveraging [Spring Data Query Methods](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.details) instead of using native SQL queries. I'll admit that query methods can get quite verbose, but on the flip side we're not mixing SQL code within Java files. Another alternative would be to write out all SQL queries in a properties file to fetch from.

Let me know if you have any questions or concerns with my comment.

felixlapierre:
Just tested and confirmed it works for both regular logs and stream logs

felixlapierre (dismissed):

sebhani:
> @sebhani I had a look at your changes. I'd like to see the stream log changes in conjunction with these changes. So, I'd prefer if you continued adding to this PR to make one PR rather than splitting the log task up. Also, I have a few details I'd like to go over.
> 
> I'd "bubble up" the request handling logic such as getting the user from the security context up to the controller. The controller is responsible for handling inputs related to the request. The security context of the request being one of those inputs.
> 
> Ideally, try to keep the DAO as light as possible such that it should simply provide a connection to the DB (Repository). However, our code-base lacks a little standard regarding the Service Layer and the DAO Layer depending on who did what. Not everyone is comfortable with layer separation. Personally, where I found that the Service layer was cut out to make the DAO a little heavier weight or the Service layer skipped over the DAO, I did not bother much about it. You can make your own judgment call on what you see.
> 
> For the Repository, look at leveraging [Spring Data Query Methods](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.details) instead of using native SQL queries. I'll admit that query methods can get quite verbose, but on the flip side we're not mixing SQL code within Java files. Another alternative would be to write out all SQL queries in a properties file to fetch from.
> 
> Let me know if you have any questions or concerns with my comment.

@MrJCipherButtles Thanks for the feedback! I have extracted the logic for passing the user id to the controller layer. Regarding the usage of  Spring Data Query Methods, I have decided not to go with these for two reasons: 1- both ways yield the same exact outcome 2- after digging into the performance of both ways, using native query is better because it doesn't involve hibernate in the process (just runs the query against our DB)

ButlerJAL:
@sebhani no worries we can keep native queries because I'm inclined to agree with you that the outcome is the same in our case. My main pet peeve is the mixing of programming languages, but I'll survive 😛.

Mind you Hibernate is pretty incredible. I never really looked deep into the actual performance factor of using a persistence framework; however, they sure make our job easier and they do come packaged with great performance enhancing features. Although I do not doubt using a persistence framework can add some overhead in certain scenarios, I believe the general sentiment should be avoided 😉.

As discussed, I'm waiting for you to review #450 because it's the older PR. I'd like to merge that first, and I anticipate there being a few conflicts, I think 2, so it really shouldn't be too bad to solve. Thank you!

felixlapierre (dismissed):

ButlerJAL (dismissed):
Different, but works great!

src/main/java/org/beanpod/switchboard/controller/LogController.java
```diff
@@ -39,17 +39,20 @@
 
   @Override
   public ResponseEntity<List<StreamLogModel>> retrieveStreamLogs(@PathVariable Long streamId) {
-    return ResponseEntity.ok(streamLogDao.getStreamLogs(streamId));
+    UserEntity user = userDao.findUser(request.getUserPrincipal().getName());
+    return ResponseEntity.ok(streamLogDao.getStreamLogs(streamId, user.getId()));
   }
```
mohhef, line 44: retrieving a stream log returns an empty list
creating a regular log results in an error for me

mohhef (changes requested):

src/main/java/org/beanpod/switchboard/controller/LogController.java
ButlerJAL: Let me know what steps you're taking because on my end I haven't encountered an error??? I'll double check

felixlapierre: Retrieving a stream log gives internal server error for me... i thought i tested it but i guess not

felixlapierre: My steps to reproduce are
- wipe the database
- compile and run the backend
- register sample sender and sample receiver
- create stream between the two devices in the frontend
- use postman to call GET /log/stream/1
```
{
    "status": "INTERNAL_SERVER_ERROR",
    "message": "could not extract ResultSet; SQL [n/a]; nested exception is org.hibernate.exception.SQLGrammarException: could not extract ResultSet",
    "error": "An unknown error occured"
}
```

ButlerJAL: I see 🤔 ... I was running the frontend and things seemed fine. I wiped the DB and tested using postman. I got this error for regular log
```
{
    "status": "INTERNAL_SERVER_ERROR",
    "message": "id is marked non-null but is null",
    "error": "An unknown error occured"
}
```

sebhani: Thanks for catching this! It's resolved right now.

felixlapierre (approved):
Can't find any errors in this one, LGTM

ButlerJAL (approved):
I've tested and LGTM. I'm not getting the internal error I reported earlier.

mohhef (approved):

PR #476, opened 2021-04-09
[#444] Make the Encoder Controller return api spec models

### LINKED ISSUE #444: All controllers should return objects using Model classes rather than objects using DTO classes

FGRCL:
# General info

This should fix the encoders returning the user's password hash

**Issue number**: #444 
**Task description**: 

 - Rewrite api spec
 - Implement the api spec in the Encoder Controller
 - Adapt the unit tests
 
# Testing

**Test coverage**:
Rewrote unit tests

# Additional notes


felixlapierre:
One minor issue of this is that when getting encoder info we return the encoder info again once per channel.

So when I get an encoder with 2 channels, I get the encoder info (and device info) three times. Once at the root level and once in each channel info.

This is a minor thing (since it's non-breaking and only a bit annoying) So i don't mind if we don't fix it but if it's an easy fix we should do it.

src/main/java/org/beanpod/switchboard/dto/mapper/DateMapper.java
```diff
@@ -0,0 +1,21 @@
+package org.beanpod.switchboard.dto.mapper;
+
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.Date;
+import org.mapstruct.Mapper;
+
+@Mapper(componentModel = "spring")
+public interface DateMapper {
```
felixlapierre, line 9: Why didn't we need this before?

felixlapierre (dismissed):

src/main/java/org/beanpod/switchboard/dto/mapper/DateMapper.java
FGRCL: Because the previously when we had only DTOs and Entities they both used `java.util.Date` for their last communications, but Open api spec uses `java.time.OffsetDatetime` for timestamps. So now that we have models we need to do the conversion between those.

mohhef: could we change the entity and dto classes to offset type and remove this class or will it have implications?

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -110,18 +111,19 @@
...
     UserEntity user = userDao.findUser(request.getUserPrincipal().getName());
 
-    Optional<EncoderDto> encoder = encoderDao.findEncoder(user, encoderDto.getSerialNumber());
-    if (encoder.isEmpty()) {
-      throw new ExceptionType.DeviceNotFoundException(encoderDto.getSerialNumber());
-    }
-    return ResponseEntity.ok(encoderDao.save(user, encoderDto));
+    return encoderDao
+        .findEncoder(user, encoderModel.getSerialNumber())
+        .map((encoderDto) -> encoderDao.save(user, encoderDto))
```
ButlerJAL: Can you please get rid of your one code smell here:
```suggestion
        .map(encoderDto -> encoderDao.save(user, encoderDto))
```

FGRCL:
> One minor issue of this is that when getting encoder info we return the encoder info again once per channel.
>
> So when I get an encoder with 2 channels, I get the encoder info (and device info) three times. Once at the root level and once in each channel info.
>
> This is a minor thing (since it's non-breaking and only a bit annoying) So i don't mind if we don't fix it but if it's an easy fix we should do it.

I'm gonna have a fix for the in the next PR that's currently in draft. it's not pretty but it works.

ButlerJAL (changes requested):
LGTM! The goal of this PR has been solved. I also see that a few extra nice additions were made such as DateFixture.java. Before approving, I'd like if you can make the simple suggested change so we do not add an extra code smell.

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
FGRCL: alright! it's fixed

ButlerJAL (dismissed):

mohhef (changes requested):

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -110,18 +111,19 @@
...
-    if (encoder.isEmpty()) {
-      throw new ExceptionType.DeviceNotFoundException(encoderDto.getSerialNumber());
-    }
-    return ResponseEntity.ok(encoderDao.save(user, encoderDto));
+    return encoderDao
+        .findEncoder(user, encoderModel.getSerialNumber())
+        .map(encoderDto -> encoderDao.save(user, encoderDto))
+        .map(encoderMapper::toEncoderModel)
+        .map(ResponseEntity::ok)
+        .orElseThrow(this::getUnknownException);
```
mohhef: getting a 400 error updating an encoder

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -76,30 +74,33 @@
...
-    return ResponseEntity.ok(encoderDao.save(user, encoderDto));
+    EncoderDto savedEncoderDto = encoderDao.save(user, encoderDto);
+    EncoderModel savedEncoderModel = encoderMapper.toEncoderModel(savedEncoderDto);
+    return ResponseEntity.ok(savedEncoderModel);
   }
 
-  @DeleteMapping("/{serialNumber}")
-  @Transactional
+  @Override
   public ResponseEntity<String> deleteEncoder(@PathVariable String serialNumber) {
```
mohhef, line 106: do you know if we were allowed to delete an encoder that has a channel? It gives an error but i think thats the expected output, just making sure.

mohhef (changes requested):

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
mohhef: image
does it have to do with my date format?

mohhef: yes it ended up being the format

mohhef: However, using the right format, the date does not get update
image

mohhef (changes requested):

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
```diff
@@ -51,17 +47,19 @@
...
+  public ResponseEntity<List<EncoderModel>> retrieveAllEncoders() {
     UserEntity user = userDao.findUser(request.getUserPrincipal().getName());
 
     List<EncoderEntity> encoderEntities = encoderDao.getEncoders(user);
     maintainDeviceStatus.maintainStatusField(encoderEntities);
-    return encoderMapper.toEncoderDtos(encoderEntities);
+    List<EncoderDto> encoderDtos = encoderMapper.toEncoderDtos(encoderEntities);
+    List<EncoderModel> encoderModels = encoderMapper.toEncoderModels(encoderDtos);
+    return ResponseEntity.ok(encoderModels);
   }
```
mohhef: i noticed when retrieving the encoders the output object has an encoder object inside it which could impact the performance tests
image
as opposed to master where it didnt have
image

not sure if that was the same issue felix had above

mohhef (changes requested):

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
mohhef: I wasnt using the right date format, but now if i try to update the date, it doesnt get changed

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
FGRCL: Yes, that is the same issue #477 will have a fix for that

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
FGRCL: That was a bug, good catch!

src/main/java/org/beanpod/switchboard/controller/EncoderController.java
FGRCL: That's a bug too, good catch! It would've been bad if that went on master.

FGRCL:
> could we change the entity and dto classes to offset type and remove this class or will it have implications?

I gave it a go, but I'm a little scared that it'll break something, so i'll make that refactor in another PR if we really want it.

ButlerJAL (approved):
Merge looks good, and PR goal is met.

mohhef (approved):
Works on my end, LGTM.
The only thing im worried is we might be adding code close the end of the project that can cause us bugs,
i think we could have solved the problem with one jsonIgnore  annotation if enforcing the api was not needed

PR #478, opened 2021-04-09
[#444] Make the Decoder Controller return api spec models

### LINKED ISSUE #444: All controllers should return objects using Model classes rather than objects using DTO classes

FGRCL:
# General info

This should fix the decoders returning the user's password hash

**Issue number**: #444 
**Task description**: 

 - Rewrite api spec
 - Implement the api spec in the Decoder Controller
 - Adapt the unit tests
 
# Testing

**Test coverage**:
Rewrote unit tests

# Additional notes
This has the changes from #476. so it'll remain as a draft until it's merged.


contracts/api-spec.yaml
```diff
@@ -268,7 +268,7 @@ paths:
         - $ref: '#/components/parameters/EncoderSerialNumber'
       responses:
         '200':
-          description: Successful retrieval of the encoder
+          description: Successful creation the encoder
```
ButlerJAL:
```suggestion
          description: Successful retrieval of the encoder
```

felixlapierre (dismissed):
All the endpoints work on my end and the code looks good to me.

ButlerJAL (commented):
LGTM other than a description not matching its operation.

Also just a suggestion if you don't mind, I've standardized Mapper methods in a previous PR, so, if you can rename the new Mapper methods you developed to fit the standard. It's basically toDto, toDtos, toEntity, toModel, toModels, etc. (No need to give descriptive names). Its implied by the Mapper class that we are doing conversion, and overloaded methods are perfect to determine the conversion based on the given parameter. It keeps code compact, uniform, and clean.

mohhef (commented):
Works good for me

FGRCL:
> Also just a suggestion if you don't mind, I've standardized Mapper methods in a previous PR, so, if you can rename the new Mapper methods you developed to fit the standard. 

Good call, I've done the changes

contracts/api-spec.yaml
ButlerJAL: I swore you changed this, but I have no way to prove it. I think your force push reversed the change
```suggestion
          description: Successful retrieval of the encoder
```

src/main/java/org/beanpod/switchboard/dto/mapper/EncoderMapper.java
```diff
@@ -16,6 +16,7 @@
     nullValuePropertyMappingStrategy = NullValuePropertyMappingStrategy.IGNORE)
 public interface EncoderMapper {
 
+  @Mapping(target = "output", qualifiedByName = "toModelsShallow")
```
ButlerJAL, line 19: Shouldn't it be toModelShallow without the 's'?

FGRCL: No, because the output is a list, so I have to specify a mapper from collections.

contracts/api-spec.yaml
FGRCL: I think merging is playing tricks on me

src/main/java/org/beanpod/switchboard/dto/mapper/EncoderMapper.java
FGRCL: No, because the output is a list, so I have to specify a mapper from collections. Also I think the conversion needs to be from a list to a set and mapstruct doesn't like that.

ButlerJAL: Gotcha!

ButlerJAL (approved):

felixlapierre (approved):

PR #500, opened 2021-04-11
Fixed an issue where CreateStreamPageContents was using callbacks instead of promises when calling DeviceApi

felixlapierre:
This issue was causing the CreateStreamPage's dialog to appear empty even when devices existed in the system.

Tests have been updated to account for this change.


gdzooks (approved):
LGTM!! :frog:

ButlerJAL (approved):
LGTM!

sharon-ho (approved):
🐸

PR #501, opened 2021-04-11
Fix issue where clicking on DeviceList breadcrumb from DeviceDetails page would break

gdzooks:
good evening, friends.

**Issue number**: :eyes:


sharon-ho (approved):
Yummy yummy in my tummy

felixlapierre (approved):

PR #507, opened 2021-04-11
Updated System Tests

ButlerJAL:
# General info

**Issue number**: n/a

**Task description**: 

- Updated SystemTests
- Added Thread.sleep to mimic more or less human speed.

# Testing

**Test coverage**:
No change

# Additional notes

**Note to reviewers**: n/a

**To do before merging**: n/a


FGRCL (approved):
This does fix the systems tests, but they seem to run slower now?

felixlapierre:
> 
> 
> This does fix the systems tests, but they seem to run slower now?

Check PR description, he added thread.sleep to mimic more human speed so we can see what's going on.

felixlapierre (approved):

ButlerJAL:
> This does fix the systems tests, but they seem to run slower now?

I added a second before every click and three second pauses when viewing a table. There are two reasons I did this, to visually see the system test run and to allow rendering.

